PR #19, opened 2026-06-03
fix(osquery): correctly detect FileVault-off without false positives (#18)

### LINKED ISSUE #18: Fix false FileVault-off alert when osquery APFS volume row disappears
## Summary

The local osquery notification pipeline can raise a **false CRITICAL “FileVault turned OFF”** alert when a single APFS/FileVault row disappears from osquery’s differential result set, even though FileVault is still enabled on the machine.

This is a security-monitoring correctness bug: it creates a high-severity false positive in the `#priority` path. The scary alert is useful when FileVault really turns off; it is actively harmful when APFS volume identity churn looks like a real protection failure. Little monitoring goblin sees one shadow move and yells “dragon.” Bad goblin.

## Repository / Component

This belongs in `webdavis/dotfiles`, not Homelab.

Likely affected files:

- `dot_local/bin/executable_osquery-results-alerter.sh`
  - Current CRIT classification lives around the `protection_off` jq function.
  - Specific risky clause currently observed:
    ```jq
    or (.name == "pack_security-policy-regression_filevault_state" and .action == "removed")
    ```
- `dot_config/osquery/packs/security-policy-regression.conf`
  - `filevault_state` currently emits one row per APFS-encrypted/FileVault-on volume.
  - `filevault_off` already exists as an absolute-state snapshot query and is probably closer to the right signal.
- Potentially any future test/fixture location if we add shell-script regression coverage.

## Incident That Exposed This

On **2026-06-02 15:07:47 MDT** / **2026-06-02 21:07:47 UTC**, osquery emitted this differential row:

```json
{
  "name": "pack_security-policy-regression_filevault_state",
  "hostIdentifier": "dresden",
  "calendarTime": "Tue Jun  2 21:07:47 2026 UTC",
  "unixTime": 1780434467,
  "epoch": 0,
  "counter": 1,
  "numerics": false,
  "columns": {
    "encrypted": "1",
    "encryption_status": "encrypted",
    "filevault_status": "on",
    "name": "/dev/disk3s1"
  },
  "action": "removed"
}
```

The alerter interpreted the removed row as:

> FileVault turned OFF

That conclusion was wrong.

## Current Behavior

`executable_osquery-results-alerter.sh` treats **any** removed row from `pack_security-policy-regression_filevault_state` as `protection_off`, which becomes `CRIT`.

Current logic excerpt:

```jq
def protection_off:
  (.name == "pack_security-policy-regression_firewall_state" and .action == "added" and (.columns.global_state // "") == "0")
  or (.name == "pack_security-policy-regression_gatekeeper_state" and .action == "added" and (.columns.assessments_enabled // "") == "0")
  or (.name == "pack_security-policy-regression_sip_state" and .action == "added" and (.columns.enabled // "") == "0")
  or (.name == "pack_security-policy-regression_screenlock_state" and .action == "added" and (.columns.enabled // "") == "0")
  or (.name == "pack_security-policy-regression_filevault_state" and .action == "removed")
  or (.action == "currently-off" and (.name | startswith("pack_security-policy-regression_")));
```

That made sense as a first approximation because the query only returns FileVault-on APFS volumes. But on modern macOS/APFS, a single row disappearing does **not** necessarily mean FileVault is off. It can mean the identity/visibility of one APFS volume changed while the actual protected Data volume remains encrypted.

## Expected Behavior

A FileVault alert should page as **CRITICAL** only when the current machine state confirms FileVault is actually off.

A single disappeared `disk_encryption` row should be treated as one of:

1. Silent/no alert, if a live verification confirms FileVault is still on.
2. NOTICE, if we want to preserve “APFS/FileVault row identity changed” as useful low-noise telemetry.
3. CRITICAL, only if verification confirms there are no FileVault-on APFS volumes or `fdesetup` reports FileVault off.

### Native FileVault status was ON

Commands run during investigation:

```bash
fdesetup status
fdesetup isactive
```

Observed result:

```text
FileVault is On.
true
```

### Native diskutil status showed FileVault still active

Relevant current facts from `diskutil`:

```text
Data volume / System data volume:
  Device Identifier: disk3s5
  Volume Name: Data
  FileVault: Yes

Mounted root snapshot:
  Device Identifier: disk3s1s1
  Volume Name: Macintosh HD
  Mount Point: /
  FileVault: Yes

Base system volume that disappeared from osquery's FileVault-on set:
  Device Identifier: disk3s1
  Volume Name: Macintosh HD
  Mounted: No
  FileVault: Yes
```

`disk3s1` is not the whole story. On sealed-system-volume macOS, the mounted root is a snapshot (`disk3s1s1`) and the user data lives on `disk3s5`. The alert fired because `/dev/disk3s1` disappeared from one osquery result set, not because the data-bearing volume became decrypted.

### Current osquery state did not say FileVault was off

Current `disk_encryption` rows included FileVault-on APFS volumes:

```json
{"encrypted":"1","encryption_status":"encrypted","filevault_status":"on","name":"/dev/disk3s1s1","type":"APFS Encryption"}
{"encrypted":"1","encryption_status":"encrypted","filevault_status":"on","name":"/dev/disk3s5","type":"APFS Encryption"}
```

The absolute-state FileVault-off check returned no rows:

```sql
SELECT 'filevault' AS protection
WHERE NOT EXISTS (
  SELECT 1
  FROM disk_encryption
  WHERE type = 'APFS Encryption'
    AND filevault_status = 'on'
);
```

No rows means: osquery’s own current-state check does **not** currently believe FileVault is off.

### No evidence of a real disable event in logs

`fdesetup` unified logs over the inspected window showed status/check calls only, not a disable/decrypt action.

Summary from the investigation:

```text
fdesetup command types in unified log, last 30d:
  51 status
   1 supportsauthrestart
   1 isUsingRecoveryKey
   1 isactive
```

`filevaultd` logs showed routine health analytics activity, not a decrypt/disable/failure trail.

### Broader posture checks did not support “active compromise caused FileVault off”

At the time of investigation:

```text
Firewall: on
Gatekeeper: on
Screen lock: on, grace=0
FileVault: on
MDM enrollment: no
Admin group: root stephen
Local Secure Token users: stephen, karl
```

Other posture notes:

- SIP was disabled, but the local osquery config already documents SIP as intentionally off on this host.
- Remote Login and File Sharing were on, but the local osquery config also documents those as currently expected/on for this host.

Conclusion from investigation: **this event was a false positive, not evidence that FileVault actually turned off.**

## Root Cause Hypothesis

The root problem is a semantic mismatch:

- `filevault_state` is a **set of rows**: “which APFS volumes currently report `filevault_status = 'on'`?”
- The notifier interprets removal of **any one row** as a boolean state transition: “FileVault turned off.”

That equivalence is invalid on modern macOS/APFS because:

- APFS exposes multiple related volumes: System, Data, Preboot, Recovery, VM, snapshots, and sometimes custom volumes like `/nix`.
- The mounted root may be a sealed system snapshot (`disk3s1s1`) rather than the base system volume (`disk3s1`).
- A row can disappear because volume identity/visibility changed, a volume became unmounted, a snapshot replaced it, or osquery’s table output changed.
- FileVault should be considered off only when the machine-level protection is off, not when one row in the volume set changes.

In short:

> “one FileVault-on APFS row disappeared” ≠ “FileVault is off.”

### Preferred behavior

Change FileVault CRIT classification from row-removal-based to verified-current-state-based.

Pseudo-logic:

```bash
if finding.name == "pack_security-policy-regression_filevault_state" \
   && finding.action == "removed"; then
  if current_state_confirms_filevault_off; then
    classify as CRIT: "FileVault turned OFF"
  else
    classify as NOTICE or suppress: "FileVault volume identity changed; FileVault still on"
  fi
fi
```

### Suggested verifier

Add a helper inside `executable_osquery-results-alerter.sh` or source it from a small helper file.

Example shape:

```bash
filevault_currently_off() {
  # Native check: authoritative and cheap.
  if command -v fdesetup >/dev/null 2>&1; then
    local status
    status=$(fdesetup status 2>/dev/null || true)
    case "$status" in
      *"FileVault is Off"*) return 0 ;;
      *"FileVault is On"*) return 1 ;;
    esac
  fi

  # osquery fallback: absolute-state check.
  local osqueryi="${OSQUERYI:-$(command -v osqueryi || echo /usr/local/bin/osqueryi)}"
  if [[ -x "$osqueryi" || -n $(command -v osqueryi || true) ]]; then
    local rows
    rows=$("$osqueryi" --json "
      SELECT 'filevault' AS protection
      WHERE NOT EXISTS (
        SELECT 1
        FROM disk_encryption
        WHERE type = 'APFS Encryption'
          AND filevault_status = 'on'
      );
    " 2>/dev/null | jq 'length' 2>/dev/null || echo 0)
    [[ "$rows" != "0" ]] && return 0
  fi

  # Fail closed or fail noisy? See decision point below.
  return 1
}
```

### Important implementation detail

Do **not** call `fdesetup`/`osqueryi` once per finding row. Cache the verifier result once per batch if any FileVault-removal row exists.

Possible approach:

1. Let jq parse rows into compact finding JSON as it does now.
2. Stop classifying `filevault_state removed` as CRIT inside the jq `protection_off` function.
3. In bash, detect whether the batch contains any FileVault row removal.
4. Run a single live verifier.
5. If verifier says FileVault is off, promote those FileVault findings to CRIT.
6. If verifier says FileVault is still on, either:
   - drop the finding entirely, or
   - downgrade it to NOTICE with explicit copy: “FileVault volume identity changed; verified FileVault still on.”

This keeps jq from needing side-effectful state checks and keeps the alert semantics clear.

### Suggested jq classification change

Current:

```jq
or (.name == "pack_security-policy-regression_filevault_state" and .action == "removed")
```

Safer first step:

```jq
# Do not classify filevault_state removed as protection_off here.
# It requires live verification because one APFS row can disappear while FileVault remains on.
```

Then add FileVault verification/promotion in bash after `raw_findings` is built.

## Alternative Implementation

Simpler, config-heavy option:

- Stop using differential `filevault_state` removal as a CRIT signal at all.
- Rely on the existing absolute-state snapshot query `filevault_off` for CRIT.
- Optionally retain `filevault_state` changes as NOTICE only.

Relevant existing config:

```json
"filevault_off": {
  "query": "SELECT 'filevault' AS protection WHERE NOT EXISTS (SELECT 1 FROM disk_encryption WHERE type = 'APFS Encryption' AND filevault_status = 'on');",
  "interval": 86400,
  "snapshot": true,
  "platform": "darwin",
  "description": "Absolute-state floor (snapshot): a row here means NO APFS volume is FileVault-encrypted, i.e. FileVault is OFF right now. Empty (silent) when on."
}
```

Tradeoff: the snapshot interval is currently 86400 seconds, so this catches current-off state only daily unless interval changes. If the goal is quick critical alerting, live verification on differential change is better.

## Decision Point: Suppress vs NOTICE

If a FileVault row disappears but verification says FileVault is still on, choose one:

### Option A — Suppress

Pros:

- Lowest noise.
- Avoids confusing “FileVault changed” messages for APFS churn.

Cons:

- Loses potentially useful forensic breadcrumb that APFS/osquery identity changed.

### Option B — NOTICE

Pros:

- Preserves a trace in `#osquery` without waking up `#priority`.
- Clearer audit trail if APFS/osquery behavior changes after OS updates.

Cons:

- Still some noise.

Recommended: **NOTICE**, at least initially, with explicit wording:

> FileVault volume identity changed — verified FileVault still on

Then demote to silent later if it proves noisy.

## Regression Test / Fixture Plan

There does not appear to be an existing test harness for these shell scripts, so add the smallest useful fixture-based check rather than a grand cathedral of test nonsense.

### Fixture 1: False positive row removal

Input line:

```json
{"name":"pack_security-policy-regression_filevault_state","hostIdentifier":"dresden","calendarTime":"Tue Jun  2 21:07:47 2026 UTC","unixTime":1780434467,"epoch":0,"counter":1,"numerics":false,"columns":{"encrypted":"1","encryption_status":"encrypted","filevault_status":"on","name":"/dev/disk3s1"},"action":"removed"}
```

Mock verifier result:

```text
FileVault is On.
```

Expected:

- No CRITICAL body is sent to `osquery-priority`.
- Either no alert is sent, or a NOTICE is sent to `osquery` saying FileVault row/volume identity changed and FileVault was verified on.

### Fixture 2: True FileVault off

Input line can be the same removed-row shape.

Mock verifier result:

```text
FileVault is Off.
```

Expected:

- CRITICAL alert is sent.
- Alert copy says FileVault actually turned OFF / verified off.
- The message includes one next step, e.g. re-enable in System Settings or investigate if user did not turn it off.

### Fixture 3: osquery absolute-off snapshot

Input line:

```json
{"snapshot":[{"protection":"filevault"}],"action":"snapshot","name":"pack_security-policy-regression_filevault_off","hostIdentifier":"dresden","calendarTime":"...","unixTime":1780434467,"epoch":0,"counter":0,"numerics":false}
```

Expected:

- CRITICAL alert still fires.
- This path should remain loud because it already represents absolute current-off state.

### Fixture 4: Non-FileVault protections still behave the same

Inputs:

- Firewall added row with `global_state = 0`
- Gatekeeper added row with `assessments_enabled = 0`
- Screen lock added row with `enabled = 0`

Expected:

- All remain CRIT as before.
- This fix must not accidentally weaken the other protection-off alerts.

## Manual Verification Commands

Useful commands for whoever implements/fixes this:

```bash
# Native FileVault state
fdesetup status
fdesetup isactive

# Current osquery view of disk encryption
osqueryi --json '
  SELECT name, type, filevault_status, encryption_status, encrypted
  FROM disk_encryption
  ORDER BY name;
'

# Current absolute-off check
osqueryi --json '
  SELECT "filevault" AS protection
  WHERE NOT EXISTS (
    SELECT 1
    FROM disk_encryption
    WHERE type = "APFS Encryption"
      AND filevault_status = "on"
  );
'

# Native APFS/FileVault detail
diskutil info /
diskutil info /System/Volumes/Data
diskutil apfs list disk3
```

## Acceptance Criteria

- [ ] A removed `pack_security-policy-regression_filevault_state` row for only `/dev/disk3s1` does **not** produce a CRITICAL alert when `fdesetup status` says FileVault is on.
- [ ] A removed FileVault row does **not** produce a CRITICAL alert when current osquery still has at least one APFS Encryption row with `filevault_status = 'on'`.
- [ ] A true FileVault-off state still produces a CRITICAL alert.
- [ ] `filevault_off` snapshot rows with non-empty snapshot still produce a CRITICAL alert.
- [ ] Firewall/Gatekeeper/SIP/screenlock CRIT behavior is unchanged.
- [ ] Alert copy distinguishes these two cases:
  - “FileVault turned OFF” = verified current protection failure.
  - “FileVault volume identity changed” = one row changed but FileVault verified still on.
- [ ] Verification work is cached once per batch, not repeated for each row.
- [ ] Add fixture/regression coverage if practical; otherwise document a manual reproduction command in the PR.

## Non-goals

- Do not patch osquery itself. Configure and interpret it correctly.
- Do not weaken real FileVault-off detection.
- Do not remove the `filevault_off` absolute-state floor; it is useful.
- Do not page `#priority` for APFS row churn unless the current protection state is actually bad.

## Final Threat Assessment From Original Incident

The original alert did **not** indicate that FileVault actually turned off. It was a false positive caused by overinterpreting one osquery differential row. No investigated evidence showed a real FileVault disable, decrypt, MDM action, or compromise path tied to that alert.

The fix is therefore about making the security monitor less stupid, not responding to an active FileVault incident.

webdavis:
Closes #18.

## The bug (two-sided)

**False positive.** The alerter flagged *any* removed `filevault_state` row as a protection-off event → a CRITICAL "FileVault turned OFF" in #priority. But `filevault_state` lists one row per FileVault-on APFS volume, and APFS volume identity churn (a base/system volume unmounting, a snapshot replacing it) drops a single row from the differential set while the data-bearing volume stays encrypted. The 2026-06-02 incident: removed `/dev/disk3s1` still reported `filevault_status: on`, `encrypted: 1` — FileVault was on.

**False negative (found while fixing the above).** Just removing the bad clause and leaning on `filevault_off` to still catch a genuine off — as both the issue and the first patch assumed — does **not** work: `filevault_off` was a **snapshot** query, and osquery writes snapshot results to `osqueryd.snapshots.log`, which the alerter never reads (it consumes `osqueryd.results.log`). So a real FileVault-off had **no path to #priority at all** — we'd have traded a false positive for a false negative on a protection turning off.

## The fix

- Drop the brittle `filevault_state … removed` clause. A removed row now falls through to a NOTICE in #osquery (breadcrumb without paging).
- Convert `filevault_off` from **snapshot** to a **differential** query (1h interval). It returns one constant row only when no APFS volume is encrypted, so its off-row is logged `added` to `results.log` when FileVault goes off (or is already off at first run) — exactly like `firewall_state`. The constant row is immune to APFS churn, so it cannot false-positive. Add the matching `protection_off` clause.

## Corrections to earlier analysis (for the record)

- The **issue description** states: *"`filevault_off` already exists as an absolute-state snapshot query and is probably closer to the right signal."* The instinct (judge absolute state, not a vanished row) was correct — but as a **snapshot** its output goes to a log the alerter never reads, so relying on it as-is is exactly what produced the false negative. This fix keeps the query's logic and changes its **mode** (snapshot → differential) so the answer lands where the consumer reads.
- An intermediate fix's commit message claimed *"genuine FileVault-off remains CRITICAL via the existing filevault_off snapshot floor … the snapshot-explode path turns that into a currently-off row"* — wrong for the same reason; that path never fires from `results.log`.

## Verification

- #18 incident row (`filevault_state` removed, `filevault_status: on`) → NOTICE in #osquery, **not** CRITICAL ✅
- `filevault_off` `added` row → CRITICAL "FileVault turned OFF" in #priority ✅
- Firewall-off and other protections → unaffected ✅
- `just l` green

## Known follow-up (separate issue)

The same snapshot-logging gap means the rest of the `*_off` "snapshot floor" (`firewall_off`, `gatekeeper_off`, `screenlock_off`) also never reaches the alerter, and the alerter's `action == "snapshot"` handling is consequently dead. Those protections are still covered by their `*_state` differential checks (and firewall/Gatekeeper by the 60s poller), so the practical gap was FileVault-only — but reviving or removing the dead snapshot floor deserves its own issue.

### DIFF
--- a/dot_config/osquery/packs/security-policy-regression.conf
+++ b/dot_config/osquery/packs/security-policy-regression.conf
@@ -59,10 +59,9 @@
     },
     "filevault_off": {
       "query": "SELECT 'filevault' AS protection WHERE NOT EXISTS (SELECT 1 FROM disk_encryption WHERE type = 'APFS Encryption' AND filevault_status = 'on');",
-      "interval": 86400,
-      "snapshot": true,
+      "interval": 3600,
       "platform": "darwin",
-      "description": "Absolute-state floor (snapshot): a row here means NO APFS volume is FileVault-encrypted, i.e. FileVault is OFF right now. Empty (silent) when on."
+      "description": "FileVault off-detector. Returns one constant row only when NO APFS volume is FileVault-encrypted (FileVault OFF), empty otherwise. Differential (NOT snapshot) on purpose: snapshot results are written to osqueryd.snapshots.log, which the alerter does not read, so as a snapshot this never reached #priority (issue #18). As a differential query the off-row is logged 'added' to osqueryd.results.log when FileVault goes off (or is already off at first run), exactly like firewall_state. The row is constant, so APFS volume identity churn cannot trigger it; it fires only on genuine FileVault-off."
     }
   }
 }
--- a/dot_local/bin/executable_osquery-results-alerter.sh
+++ b/dot_local/bin/executable_osquery-results-alerter.sh
@@ -73,17 +73,31 @@ raw_findings=$(printf '%s\n' "$new_lines" | jq -rR '
   . as $line | (try ($line | fromjson) catch empty) |
   # A security-policy row is CRITICAL only when the protection turned OFF, not
   # on every change. For the boolean states that is an "added" row carrying the
-  # off value; for filevault (the query returns only encrypted volumes) it is a
-  # "removed" row — a volume left the encrypted set. Re-enables, version bumps,
-  # sharing changes, and the paired "removed" old-value rows fall through to
-  # NOTICE. (sharing cannot be direction-classified from a single row, so it is
-  # always NOTICE; the poller covers firewall/Gatekeeper transitions too.)
+  # off value. Re-enables, version bumps, sharing changes, and the paired
+  # "removed" old-value rows fall through to NOTICE. (sharing cannot be
+  # direction-classified from a single row, so it is always NOTICE; the poller
+  # covers firewall/Gatekeeper transitions too.)
+  #
+  # FileVault is NOT detected via filevault_state. That query emits one row per
+  # FileVault-on APFS volume, and on sealed-system-volume macOS a single row can
+  # leave the differential set (a base system volume unmounting, a snapshot
+  # replacing it, APFS identity churn) while the data-bearing volume stays
+  # encrypted — so "one filevault_state row removed" does NOT mean FileVault is
+  # off (issue #18). A removed filevault_state row falls through to NOTICE.
+  #
+  # Genuine FileVault-off is detected by the filevault_off query, which emits a
+  # single constant row only when NO APFS volume is encrypted. It is a
+  # DIFFERENTIAL query (not snapshot): snapshot output goes to
+  # osqueryd.snapshots.log, which this alerter does not read, so the earlier
+  # snapshot form never reached here (the false-negative half of issue #18). As
+  # a differential query its off-row is logged "added" to results.log, matched
+  # below. The constant row is immune to APFS churn, so no false positive.
   def protection_off:
     (.name == "pack_security-policy-regression_firewall_state" and .action == "added" and (.columns.global_state // "") == "0")
     or (.name == "pack_security-policy-regression_gatekeeper_state" and .action == "added" and (.columns.assessments_enabled // "") == "0")
     or (.name == "pack_security-policy-regression_sip_state" and .action == "added" and (.columns.enabled // "") == "0")
     or (.name == "pack_security-policy-regression_screenlock_state" and .action == "added" and (.columns.enabled // "") == "0")
-    or (.name == "pack_security-policy-regression_filevault_state" and .action == "removed")
+    or (.name == "pack_security-policy-regression_filevault_off" and .action == "added")
     or (.action == "currently-off" and (.name | startswith("pack_security-policy-regression_")));
   def sev:
     if protection_off


webdavis:
Superseded and discarded without merging. The FileVault-off fix (both the false-positive and false-negative halves of #18) is folded into the `feat/osquery-alerter-three-tier` branch, whose commit carries a `Closes #18` trailer. That branch reworks filevault_off as a differential query (not snapshot) and removes the dead absolute-state floor — a strictly larger fix than this PR. Closing as dead.

PR #31, opened 2026-06-23
herdr migration, headless Tailscale, weekly Homebrew, and moshi notifications

webdavis:
## What this does

Five mostly-independent pieces of local-environment work, plus the specs behind them.

### Shell startup

Non-interactive login shells (`bash -lc`, `ssh host cmd`) were paying for interactive-only setup they never use. That setup now sits behind a single interactive guard, which drops about 85ms off those shells. The other big cost was Homebrew: instead of running `brew shellenv` on every shell (~27ms), `~/.bashrc` reads a cache that gets regenerated from the real generator at apply time. A test asserts the cache matches live output, and `~/.bashrc` regenerates it in the background if it drifts. `~/.profile` now only sources `~/.bashrc` for interactive shells.

### tmux to herdr

tmux, sesh, and the tmux2k status hacks are gone; herdr's config is tracked instead. The notable bits:

- Eight workspace quick-jump chords, agent-pane navigation, and a bounce back to the last workspace.
- Ctrl-h/j/k/l moves between Neovim splits and herdr panes without caring which is which. It went through three forms before settling: a shell script, then a Rust binary, now a herdr plugin built and linked on apply.
- A second Rust plugin handles the last-workspace bounce.
- Worktrunk now creates worktrees where herdr expects them, so the two tools agree.
- A fresh terminal lands in the homelab workspace.

### Homebrew upgrades

Upgrades were running daily and unattended, so apps restarted and permission prompts fired at random times, including when nobody was at the machine. They now run on a LaunchAgent every Monday at noon, when someone's present. The daily job only refreshes metadata. The upgrade helper keeps going if a step fails and logs what broke, with a test covering that.

### Tailscale, headless

Moved off the GUI app to the open-source formula, installed as a launchd system daemon. It comes up before login and doesn't need a permission re-approval click after updates, which matters for a machine I'm not always sitting at. The weekly Homebrew job re-installs the daemon binary when the formula upgrades. A reminder script (no sudo) prints the one-time auth steps if the daemon is ever down or signed out.

### moshi (phone notifications)

moshi bridges agent sessions to a phone app. Install and pairing are declarative now. When a Claude Code session finishes, a Stop hook sends a push with the project, git branch, and a short snippet of the last response. The secret for that push is read at runtime from a 0600 file rendered from the password manager, so it never lands on a command line or in process arguments.

### Packages, skills, and settings

- `codex` and `codex-app` casks replace the npm `@openai/codex`; the `paseo` cask is now tracked.
- Vendored the moshi and herdr agent skills with a recipe to refresh them, and tracked the agent skill files plus `AGENTS.md`.
- Claude Code keeps its policy fields (permissions, hooks, plugins, status line) managed while leaving ergonomic toggles alone. Default effort is pinned and the rust-analyzer plugin is on.

## Notes for review

- Specs and plans for the larger pieces are under `docs/superpowers/`, and the commit history is grouped by area if you want per-change detail.
- Every commit passes the pre-commit gate (lint + tests); the suite is green on the tip.
- Nothing sensitive is committed. The push secret and Tailscale auth come from the password manager at apply or runtime, not the repo.


webdavis:
Closed by the D1 cutover: main now carries this work through the merged slice PRs, and the converged topology soaked before closure.

PR #32, opened 2026-07-04
DO NOT MERGE — integration reference (modernization)

webdavis:
**DO NOT MERGE.** This PR is the frozen live-state reference for the repo-modernization effort — the
exact combined diff (PR #31 + PR #25 + the SP1 age-encryption work) that dresden runs today.

It is **never** merged. Its role is a reviewable snapshot while the same work is reimplemented from
`main` as small, self-contained PRs (see `docs/superpowers/plans/2026-07-03-sp2-combine-and-split.md`).
Each small PR fully wires one feature and is reviewed on its own; this branch is closed at cutover with
pointers to the landed slices.

Freeze policy: no new feature work lands here; only hotfixes needed to keep dresden healthy, and every
such hotfix is also folded into its corresponding reimplementation slice.

PR #33, opened 2026-07-04
docs: import modernization specs and plans (SP2 slice S1)

webdavis:
**SP2 slice S1 of 12** — first reimplementation slice of the modernization effort (reference: #32, plan:
`docs/superpowers/plans/2026-07-03-sp2-combine-and-split.md`).

## What this slice ships

- The 21 modernization docs (`docs/plans/` brief + `docs/superpowers/{plans,specs}/`) imported verbatim
  from the frozen `integration/modernization` branch — these are the specs and plans every later slice
  PR references.
- `AGENTS.md` as a symlink to `CLAUDE.md`, plus its `.chezmoiignore` line so chezmoi never deploys it to
  `$HOME`. That ignore line is this slice's only hunk of the shared `.chezmoiignore` — verified that the
  remaining diff vs the integration branch contains only other slices' hunks.

## Ledger fixes folded in

None — S1 has no ledger fixes assigned.

## Operator apply needed

No. Docs are `.chezmoiignore`d (reference-only) and the new `AGENTS.md` entry only tells chezmoi to keep
ignoring a repo-root file. Nothing reaches `$HOME`.

## Verification

- `git diff slice/docs integration/modernization -- docs/ AGENTS.md` — empty (byte-identical import).
- `AGENTS.md` git mode `120000`, resolves to `CLAUDE.md`.
- `just lint-check` green (independently re-run during task review).

Merge convention: `--merge` (never squash), subject `Merge pull request #N from webdavis/slice/docs (#N)`.


AGENTS.md
```diff
@@ -0,0 +1 @@
+CLAUDE.md
```
webdavis, line 1: Does this need to have a leading `@`, as in `@CLAUDE.md`?

webdavis: No — it must be the literal filename. Two different mechanisms are in play:

1. **This is a symlink, not file content.** The blob is the 9-byte target path `CLAUDE.md` (what you see on line 1 is how git stores a symlink). A target of `@CLAUDE.md` would point at a nonexistent file named `@CLAUDE.md` — a dangling link, and every tool reading `AGENTS.md` would get nothing. Verified locally: `ls -l AGENTS.md` → `AGENTS.md -> CLAUDE.md`, resolves to the full file.

2. **`@file` is Claude Code's import syntax, which lives *inside* a memory file's body** (e.g. a `CLAUDE.md` containing a line `@README.md`). It would only apply if `AGENTS.md` were a regular file whose body imports `CLAUDE.md` — but then only Claude Code would resolve the import; Codex and other agents.md-convention tools read plain markdown and would see a literal one-line file. The symlink does the indirection at the filesystem level, so every tool gets the full content with no syntax support needed.

PR #34, opened 2026-07-04
throwaway: S2 CI red-verification (do not merge)

webdavis:
Temporary branch to verify the hardened CI workflow gates correctly (green at slice head, then red with deliberate breakage). Will be closed and the branch deleted.

PR #35, opened 2026-07-04
feat(lint): treefmt-nix orchestrator, test harness, and hardened CI (SP2 slice S2)

webdavis:
**SP2 slice S2 of 12** — the checkable foundation (reference: #32, plan:
`docs/superpowers/plans/2026-07-03-sp2-combine-and-split.md`, research amendments R2 + R3).

## What this slice ships

- **treefmt-nix replaces `scripts/lint.sh`** (493 lines deleted) as the lint/format orchestrator:
  flake-wired (`treefmt-nix.lib.evalModule`), one global excludes list replaces the 6×-duplicated prune
  set, `nix flake check` is now the drift gate. Style preserved exactly: `shfmt -i 2 -ci -s`, mdformat
  105 cols, `nixfmt-rfc-style` (the bare `nixfmt` attr in nixpkgs 25.05 is classic — pinned the RFC
  style lint.sh actually used), taplo with `dot_aerospace.toml` excluded, jq, yq.
- **lint.sh's chezmoi-specific checks ported as custom treefmt formatters**: shellcheck-on-rendered
  templates (`CI=1 chezmoi execute-template`), osquery config render + jq validation. Build-time
  shellchecked via `writeShellApplication`; they run inside the Nix check sandbox on every system, which
  the old macos-only CI never did.
- **Test harness introduced to main**: `just test` runs every executable `test/*.sh` (green on the
  still-empty tree — real tests arrive with their feature slices) plus bats suites inside
  `nix develop .#run`.
- **Pre-commit hook now three stages, fail-closed**: lint check → `just test` →
  `gitleaks git --staged --no-banner --redact` (gitleaks already in the brew manifest on main).
- **CI hardening (R3)**: actions SHA-pinned with version comments (`actions/checkout@9c091bb` = v7.0.0;
  `NixOS/nix-installer-action@6b8548fe` = head of main, the action publishes no tags); workflow-level
  `permissions: contents: read`; `persist-credentials: false`; CI runs the flake drift gate + `just
  test` + zizmor (offline); actionlint via its treefmt module; `.github/dependabot.yml`
  (github-actions, weekly, review-gated).

## Ledger fixes folded in

- CI now actually runs tests and fails on drift — **empirically verified**: green control run
  [28720500730](https://github.com/webdavis/dotfiles/actions/runs/28720500730) at this branch's head,
  then deliberate-breakage run [28720576155](https://github.com/webdavis/dotfiles/actions/runs/28720576155)
  **failed at the drift gate** (throwaway branch since deleted; temporary draft PR #34 existed only to
  trigger the workflow and is closed).
- SHA-pin instruction upgraded per R3 (the installer action has zero tags/releases, so `@vX` was
  impossible).
- lint.sh's runner-selection subshell bug and prune-set duplication die with the file.
- Deferred out (conductor decisions): the bats `lib.bash` zero-count fix ships with its file in S9;
  `test-brew-cache`/`brew-upgrade` recipes belong to S11/S6.

## Intentional divergences from the integration branch

This slice is a redesign (R2): the integration branch's lint.sh-based justfile/CI/hook hunks are
superseded rather than carried. Recorded for the Phase D reconciliation gate.

## Operator apply needed

No. Everything here is source-only (`.chezmoiignore`d) — nothing reaches `$HOME`.

## Dependabot auto-merge (added at review, operator-approved)

Per review discussion: Dependabot PRs now **auto-merge themselves** once the required `lint` check
passes — you only hear about a bump when CI rejects it.

- `dependabot.yml` gains `cooldown: default-days: 7` — Dependabot never proposes a release younger than
  a week (the window in which upstream compromises historically get caught and yanked).
- New `dependabot-automerge.yml`: actor-guarded to `dependabot[bot]`, workflow-level `contents: read`
  with job-level elevation, `dependabot/fetch-metadata` SHA-pinned (`25dd0e3` = v3.1.0, verified against
  the tag ref), merges via the house `--merge` method.
- Repo-side (already applied via API): **Allow auto-merge** enabled; **branch protection on `main`**
  requires the `lint` check (admins exempt, so direct pushes by the operator are unaffected).

Merge convention: `--merge`, subject `Merge pull request #N from webdavis/slice/lint-ci (#N)`.


.github/dependabot.yml
webdavis, line 1: - What does dependabot do here?
- Not sure I want to review these at all. It depends what it is. Explain.

webdavis: **What it does here:** with `package-ecosystem: github-actions`, Dependabot watches exactly one thing — the two `uses:` lines in `.github/workflows/lint.yml` (`actions/checkout@<sha>` and `NixOS/nix-installer-action@<sha>`). Once a week it checks whether upstream published something newer; if so, it opens a PR that bumps the pinned SHA and the `# vX.Y.Z` comment. That's the whole surface. It never touches brew, nix, npm, or anything else — no other ecosystem is configured.

**Why it exists at all:** this PR pins both actions to immutable commit SHAs (the supply-chain fix — a tag like `@v7` can be silently re-pointed at malicious code, which is how the tj-actions compromise propagated; a SHA cannot). The trade-off of pinning is that the pin never moves, so you'd stop getting upstream security/bug fixes forever. Dependabot is just the freshness mechanism for the pins: review-gated, no auto-merge, nothing changes unless you merge it.

**What reviewing one costs:** the diff is one line per action (old SHA → new SHA + new version comment), with release notes linked in the PR body. CI runs on the bump PR itself, so the drift gate + tests validate it before you look. Realistically ~30 seconds, a few times a year — `actions/checkout` releases infrequently, and the nix installer moves slowly.

**If you'd rather not see these at all,** that's a legitimate choice: delete this file and the pins simply stay frozen. A stale SHA is still *tamper-proof* (immutability is the security property); the only risk is staleness. Middle ground: change `interval: weekly` to `monthly`. My recommendation is keep it as-is — it's the lowest-effort way to keep the pins from rotting, and silence means no updates exist.

PR #37, opened 2026-07-09
feat(herdr): migrate tmux/sesh to herdr atomically (S4)

webdavis:
Ships the tmux→herdr migration as one atomic slice (S4 of the SP2 combine-and-split plan): main never
holds both multiplexers, nor neither.

## What this ships

- **herdr arrives:** `dot_config/herdr/config.toml` (workspaces, jump chords, plugin_action
  keybindings), the two Rust plugins `herdr-last-workspace` + `herdr-smart-nav` (transplanted **with
  their test suites** — 17 `cargo test`s green), the curl-based installer
  (`run_onchange_before_15-install-herdr`), the two plugin build scripts, and
  `~/.local/bin/herdr-jump.sh`.
- **tmux/sesh dies atomically:** `dot_tmux.conf`, `dot_config/sesh/**`, the six `tmux-*`/`sesh-*`
  helper scripts, the tmux2k installer script — plus `claude-restart.sh` **and its
  `com.claude.code.plist` LaunchAgent + loader** (moved here from S7: the plist's only payload execs the
  tmux-coupled script, and the loader's chezmoi `include` of the deleted plist would have broken every
  future `chezmoi apply`).
- **Shared-file hunks (S4's only):** bashrc swaps TERM/aliases/notifier-skip/auto-attach to herdr
  (relay + shell-foundation hunks left for S7/S11); CLAUDE.md gains Herdr Workspace Management, Herdr
  Native Status, and smart-nav docs; the global toolchain line now says herdr; `sesh`+`tmux` leave the
  brew manifest; Rust `target/` ignore rules added.

## Ledger fixes folded in (all test-driven where they have runtime surface)

- The two plugin build scripts consolidated into a shared `.chezmoitemplates/herdr-plugin-build.sh.tmpl`
  partial, and the plugin-link check's `grep -q "$plugin_id"` **anchored** (`^- ${plugin_id} `, decided
  from live `herdr plugin list` output) — red→green via `test/herdr-plugin-build-template.sh`.
- `dot_fzf_bindings`: 14 dead tmux-guarded widget branches removed.
- `dot_bash_bindings`: the duplicate `\C-gss` vi-command binding fixed (`\C-gsh`, matching its
  insert-mode sibling) — red→green via a new `test/bash-bindings-unique.sh` invariant (no
  `(keymap, key-seq)` bound twice; 279 bind lines covered); stale `nvm` bindings dropped; the `$blue`
  audit found and fixed **two real defects** (a use-before-definition and an unscoped global leak that
  masked it).
- The three new herdr chezmoiscripts added to treefmt's rendered-shellcheck gate.

## Operator apply required (KeePassXC not needed, but live-state changes are real)

`chezmoi apply` from an interactive terminal after merge will: install herdr via the official curl
installer (+ preview channel), **build the two Rust plugins at apply time** (cargo required), deploy the
herdr config + bashrc swap, and remove the deployed tmux/sesh files. `brew bundle --cleanup` will
uninstall `tmux` and `sesh` on the next manifest-triggered run. The `com.claude.code` LaunchAgent is no
longer managed; `launchctl bootout gui/$UID/com.claude.code` once to retire the live registration.
Reminder (two-world trap): dresden's live source is still `integration/modernization` — do not apply
from there expecting this PR's state; live convergence is D1's job.

## Verification

- Full gate green on every commit (treefmt lint-check incl. rendered-template shellcheck, `just test`
  incl. the two new red→green tests, gitleaks).
- P-4 sweeps: zero tmux/sesh references left outside justified keeps (an "≈ tmux sessions" analogy, the
  migration-spec filename, generic bat comments, the moshi fork's SKILL.md — banked as a Phase E item);
  every deleted basename referenced nowhere; every `config.toml` consumer exists.
- Shared-file assertion: this branch's remaining diff vs `integration/modernization` on the shared
  files contains zero tmux/sesh/herdr hunks — S4 took exactly its own.


.chezmoiscripts/run_onchange_after_55-build-herdr-last-workspace-plugin.sh.tmpl
```diff
@@ -0,0 +1,36 @@
...
+  if focused=$(herdr workspace list 2>/dev/null |
+    jq -r '[.result.workspaces[] | select(.focused) | .workspace_id][0] // empty' 2>/dev/null) &&
+    [[ -n $focused && $focused != "null" ]]; then
+    mkdir -p "$(dirname "$state")"
+    printf '%s\n\n' "$focused" >"$state"
+  else
+    echo "could not seed last-workspace MRU (herdr server down?); it self-seeds on first workspace focus" >&2
+  fi
+fi
+{{ end -}}
```
webdavis, line 29: If seeding is part of the plugin, then why isn't this written in rust like the plugin so that it's language consistent?

.chezmoitemplates/herdr-plugin-build.sh.tmpl
```diff
@@ -0,0 +1,52 @@
...
+
+# Resolve cargo. On a fresh machine, run_once_before_20 installs rustup into
+# ~/.cargo/bin during THIS same apply, but that dir is not on the apply shell's
+# PATH yet — so fall back to the absolute path before giving up. A missing
+# toolchain must NEVER abort the whole apply (this partial is includeTemplate'd
+# into run_onchange chezmoiscripts, so exit 1 here kills the entire apply).
+if command -v cargo &>/dev/null; then
+  cargo_bin="cargo"
+elif [[ -x "$HOME/.cargo/bin/cargo" ]]; then
+  cargo_bin="$HOME/.cargo/bin/cargo"
```
webdavis: Shouldn't the cargo binary PATH be absolute so that the script is deterministic. If cargo is in a different place than expected then it's symptomatic of an incorrect expectation of where cargo is installed.

dot_config/herdr/config.toml
```diff
@@ -0,0 +1,219 @@
...
+key = "ctrl+k"
+type = "plugin_action"
+command = "herdr-smart-nav.nav_up"
+description = "nav up: Neovim split or herdr pane"
+
+[[keys.command]]
+key = "ctrl+l"
+type = "plugin_action"
+command = "herdr-smart-nav.nav_right"
+description = "nav right: Neovim split or herdr pane"
```
webdavis, line 206: Why is the command execution using "dot" calls? For example: `herdr-smart-nav.nav_right`, instead of `herdr-smart-nav --nav right` like a standard command line invocation?

dot_local/share/herdr/plugins/herdr-last-workspace/src/main.rs
```diff
@@ -0,0 +1,168 @@
...
+        let ev = r#"{"event":"workspace_focused","data":{"type":"workspace_focused","workspace_id":"w18"}}"#;
+        assert_eq!(parse_focused_id(ev), Some("w18".into()));
+    }
+
+    #[test]
+    fn parse_focused_id_handles_garbage() {
+        assert_eq!(parse_focused_id("not json"), None);
+        assert_eq!(parse_focused_id("{}"), None);
+    }
+}
```
webdavis, line 319: Is it standard practice in rust to include the unit tests in the same file as the production code?

dot_local/share/herdr/plugins/herdr-smart-nav/src/main.rs
```diff
@@ -0,0 +1,196 @@
...
+        env::var("HERDR_ACTIVE_PANE_ID").ok(),
+    );
+    let is_nvim = match pane.as_deref() {
+        Some(p) => is_nvim_foreground(&herdr_capture(&["pane", "process-info", "--pane", p])),
+        None => false,
+    };
+    execute(&decide(pane.as_deref(), is_nvim, &direction, chord));
+}
+
+#[cfg(test)]
```
webdavis, line 106: Same as other rust file. Not sure it's good practice to keep unit tests in production code file.

dot_bashrc.tmpl
```diff
@@ -268,7 +265,7 @@ __cmd_notify_precmd() {
   local elapsed=$((SECONDS - __cmd_notify_start))
   __cmd_notify_start=""
   # Skip known interactive TUIs.
-  [[ $__cmd_notify_name =~ ^(vim|nvim|less|man|top|btop|ssh|tmux|claude|fzf) ]] && return
+  [[ $__cmd_notify_name =~ ^(vim|nvim|less|man|top|btop|ssh|herdr|claude|fzf) ]] && return
```
webdavis: Add `hermes` and `codex` to this list.

webdavis (commented):
Implement the fixes, and answer the questions.

dot_config/herdr/config.toml
webdavis: The dot form is not our CLI syntax — it is herdr's binding grammar for `type = "plugin_action"`: `command = "<plugin-id>.<action>"` names an **action declared in the plugin's `herdr-plugin.toml` manifest**, and herdr invokes the plugin's registered handler directly as argv (no shell). A `herdr-smart-nav --nav right` invocation would require `type = "command"`, which herdr wraps in `/bin/sh -lc` — exactly the wrapper the plugin migration removed (see `docs/superpowers/specs/2026-06-23-herdr-nav-plugin-action-design.md`: plugin_action is the idiomatic integration; the argv/flag form is what we migrated away from).

dot_local/share/herdr/plugins/herdr-last-workspace/src/main.rs
webdavis: Yes — this is the idiomatic Rust convention, not a shortcut. The Rust Book's test-organization chapter (https://doc.rust-lang.org/book/ch11-03-test-organization.html) prescribes exactly this: **unit tests live in a `#[cfg(test)] mod tests` block in the same file as the code they test** (same-file placement is what allows testing private functions), while *integration* tests go in a separate `tests/` directory. The `#[cfg(test)]` attribute compiles the module out of every non-test build, so `cargo build --release` ships zero test code.

dot_local/share/herdr/plugins/herdr-smart-nav/src/main.rs
webdavis: Same answer as the other plugin: in-file `#[cfg(test)] mod tests` is the Rust Book's prescribed home for unit tests (ch. 11.3) — it enables private-function testing and is compiled out of release binaries entirely. Moving these to `tests/` would also change their nature: integration tests only see the public interface, and these are **bin** crates — `tests/` can't link a binary's internals without restructuring into a lib+bin split, real cost for zero gain here.

.chezmoiscripts/run_onchange_after_55-build-herdr-last-workspace-plugin.sh.tmpl
webdavis: Implemented in `e56efaa` — you were right, and it came out better than language consistency alone: the seed is now a `seed` subcommand ON the plugin itself, resolving the state path via the plugin's own `state_file()` and querying herdr the same way `bounce` does (the bash/jq seed is gone entirely). Seeder/reader path divergence is now structurally impossible. Red→green: 6→14 `cargo test`s (seed decision, idempotence, `HERDR_PLUGIN_STATE_DIR` override, cold-start), plus end-to-end on the real binary (cold-start seeds; re-run is a no-op; server-down exits 0 with a hint). Argv-collision analysis: herdr dispatches only the manifest-declared `record`/`bounce`; `seed` is invoked solely by the chezmoiscript — purely additive match arm.

.chezmoitemplates/herdr-plugin-build.sh.tmpl
webdavis: Implemented in `e63e2d6` — accepted your reasoning verbatim: the PATH probe encoded a fuzzy expectation, and a cargo found elsewhere would be a symptom, not a convenience. The partial now resolves `"$HOME/.cargo/bin/cargo"` (rustup's canonical location, provisioned by run_once_before_20) absolutely and ONLY; absent → skip-with-hint naming rustup as the expected provisioner. Red→green: new test case asserts a PATH-only cargo is now ignored (failed against the old partial, passes now).

dot_bashrc.tmpl
webdavis: Implemented in `e993cf1` — `hermes` and `codex` added (this deliberately overrides S4's earlier "codex rides with S7" scoping, per your order; noted in the commit body). One behavior note, pre-existing to the whole list: the regex is prefix-anchored, so `hermes` also skips e.g. `hermesx` — consistent with how `top`/`ssh`/`claude` already behave. Verified: `hermes gateway restart` and `codex` skip; `cargo build` still notifies.

PR #38, opened 2026-07-10
feat(tailscale): headless daemon — formula swap, status reminder (S5)

webdavis:
Ships the Tailscale headless-daemon slice (S5 of the SP2 combine-and-split plan).

## What this ships

- **Atomic manifest swap:** `tailscale-app` GUI cask out, `tailscale` formula in (one commit — main
  never holds both, nor neither). Correctly alphabetized (the reference branch's copy is misplaced;
  D1's empty-diff gate will show that as an intentional improvement).
- **`run_onchange_after_66-tailscaled-status.sh.tmpl`** — a sudo-free status reminder (byte-identical
  transplant from the reference branch): when the daemon is down or unauthenticated it prints the
  one-time manual steps; it never runs sudo, never authenticates, and every path exits 0 (verified
  against its code by independent review).
- **treefmt render-lint coverage** for the new template (per the Phase E `fix/template-render-coverage`
  rule: every slice adds its own templates).
- **CLAUDE.md "Tailscale (headless daemon)" section** — with its Updates paragraph deliberately
  describing the MANUAL `sudo tailscaled install-system-daemon` re-copy: the weekly automation lives in
  `homebrew-weekly-upgrade.sh`, an S6 file, and this repo does not document code it doesn't carry (S6
  restores the automated wording). Machine-specific claims (passwordless sudo) hedged as
  dresden-notes.
- **Collateral (review-surfaced):** `test/update-skills-launchagent-path.sh` made hermetic — it called
  bare `chezmoi execute-template`, which reads the machine's live sourceDir and flakes whenever that
  checkout is on another branch; now pinned via `--source` like treefmt's render check.

## Re-scope notes (per the plan's S5 section, updated in this PR)

The four osquery-tailscale-monitor files in the reference delta (loader, plist, poller script, bats)
are **S9's** — they're osquery components and carry the `2f430b3` fix with them. A tailscale-named
file is not automatically S5's.

## Operator notes

- **No apply needed now** — dresden already runs the headless daemon live; convergence is D1's job
  (two-world model).
- The one-time `sudo tailscale up --accept-dns=true` + **Disable Key Expiry** remain manual and are
  documented in the status script (not automatable by design).

## Verification

Full gate green on every commit (treefmt incl. rendered-shellcheck of the new template, `just test`,
gitleaks). Independent whole-branch review: 5 findings (2 Important, 3 Minor — doc-accuracy + a stray
uncommitted local edit, operator-confirmed external) → all fixed → delta re-review **CLEAN**.


test/etc-hosts-fallback-record.sh
```diff
@@ -0,0 +1,77 @@
...
+#   1. the record exists (description mentions MagicDNS) and has sudo: true
+#   2. the command is `sh -c '...'`-wrapped (redirect lands as root under the runner)
+#   3. run once against a temp hosts file -> the tab-separated pin line is appended
+#   4. run twice -> file byte-identical (idempotent guard-then-append)
+set -euo pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+YAML="$REPO_ROOT/.chezmoidata/macos_system_setup.yaml"
+PIN_NAME="mister.tail2f2430.ts.net"
+PIN_LINE=$'100.109.58.54\tmister.tail2f2430.ts.net\tmister'
```
webdavis: We shouldn't hardcode data like this; it should be kept in a yaml file in `.chezmoidata/` and read from.

Follow this rule for any other mentions for similar data.

webdavis (commented):
Only one fix that I can see. Feel free to push back if you need to.

test/etc-hosts-fallback-record.sh
webdavis: Implemented in `c90a700` — and taken as the standing rule you stated: the test now derives EVERY expectation by parsing the `.chezmoidata` record itself (yq extract → `[[ =~ ]]` parse of the printf format → cross-check that the grep guard watches the same FQDN the printf appends → CGNAT shape check so a mangled record fails loudly instead of passing vacuously). Zero tailnet constants remain anywhere outside the YAML record — swept the whole branch (`git grep -E '100\.109|tail2f2430' -- ':!.chezmoidata'` → no hits; CLAUDE.md was already mechanism-only). Independent review re-ran the mutations (IP→8.8.8.8 fails; guard/printf mismatch fails; faithful record green): CLEAN. One cosmetic note it left, non-blocking: the CGNAT shape regex is looser (100/8) than its comment claims (100.64/10) — say the word if you want it tightened to match, otherwise it stands as a heuristic.

PR #53, opened 2026-07-12
feat: S10 macOS defaults tooling + SSH hardening + R8 endpoint casks

webdavis:
## Context

Main is being rebuilt slice by slice while the live machine (dresden) applies from `integration/modernization` until the D1 cutover. S10 is the macOS defaults / system-setup slice: it fixes the defaults-management tooling, closes a real SSH password-login hole, carries the already-approved endpoint-hardening casks to main, and adds a check-only security-posture reminder.

You asked to review this PR yourself before merge, so it is open, not auto-merged. It touches SSH hardening on the machine you connect to, so it went through the full three-pass adversarial review (two dual rounds plus an isolated final pass), with the sshd component attacked hardest.

## Summary

- The macOS-defaults tools now resolve their source per-worktree, fixing a worktree-writes-primary bug.
- The SSH drop-in closes the PAM/keyboard-interactive password hole and is wired for fresh machines.
- The drop-in is renamed to win include-precedence, and its verifier now catches Match-block bypasses.
- LuLu + OverSight casks and firewall stealth mode reach main (the R8 endpoint hardening).
- A check-only FileVault/Gatekeeper/SIP posture reminder reports on every apply, never changes state.

Key files:
- `dot_local/bin/executable_ssh-hardening.sh`
- `dot_local/bin/macos-defaults-lib.sh` + the defaults trio
- `.chezmoiscripts/run_after_67-macos-security-posture.sh.tmpl`
- `.chezmoidata/macos_system_setup.yaml`, `.chezmoidata/system_packages_autoinstall.yaml`

## What changed

- Defaults tooling: the apply/capture/drift trio share a lib that resolves the chezmoi source via `chezmoi source-path`, so running a capture from a secondary worktree writes that worktree's YAML, not the primary checkout. The `after_41` sudo runner is guarded with `{{ if index . "sudo" }}` so a record without a sudo key no longer throws.
- SSH hardening: the drop-in sets `PasswordAuthentication no`, `KbdInteractiveAuthentication no`, `UsePAM yes`, `PubkeyAuthentication yes`, `PermitRootLogin no` (the audit's accepted effective config; the old drop-in set only the first, leaving the keyboard-interactive/PAM password path open). It is renamed `000-ssh-hardening.conf` so it wins first-value-wins over Apple's `100-macos.conf`, migrates the old `50-` file away, and is wired as a `macos_system_setup.yaml` record so a fresh machine actually locks sshd. The verifier checks three ways (global `sshd -G`, a raw scan of every included file for a `Match` block re-enabling a protected directive, and `sshd -G -T -C` for a root and a normal-user connection spec), and fails closed when it cannot run. The operator-only `--reload` validates the full config and probes real sshd readiness before trusting a restart.
- Endpoint hardening (R8, previously operator-approved): LuLu (outbound firewall) and OverSight (mic/camera alerts) casks, plus a firewall record that enables the Application Firewall and stealth mode.
- Posture reminder: reports FileVault / Gatekeeper / SIP on every apply; a failed query reads "could not determine" rather than a false "enabled"; it never enables anything.

## How it was reviewed

- Three sequential adversarial passes (two dual Opus+sol rounds, one isolated sol final gate), sol hardened each round. Round 1 caught the include-precedence hole and that the local de-homebrewed gate was itself invalid; round 2 caught that the verifier was blind to `Match`-block re-enables and that the post-reload check proved job-load not readiness; round 3 found nothing new.
- Conductor verification independently reproduced the `Match Group` bypass against the fixed verifier (correctly rejected, including case/tab variants) and ran the corrected de-homebrewed gate.
- 13 findings fixed red-first across the two rounds; the bats suite plus new SSH/posture/defaults tests all pass or skip cleanly.

## Operator steps (this slice needs you, physically present)

1. Apply from a LOCAL console (not over the SSH session being hardened), KeePassXC unlocked. The Tier-2 runner writes `000-ssh-hardening.conf`, removes the old `50-`, enables the firewall + stealth, and installs the casks. One sudo prompt.
2. Live sshd reload + proof (deliberately not automated): keep the current session open, run `~/.local/bin/ssh-hardening.sh --reload` (it validates and probes readiness), then prove a key-only login in a NEW session before closing the old one. Verify with connection specs (the plain `sudo sshd -T` shares a Match blind spot): `sudo /usr/sbin/sshd -G -T -C user=root,addr=203.0.113.1,host=localhost | grep -iE 'passwordauthentication|kbdinteractive|usepam|pubkey|permitrootlogin'` reads no/no/yes/yes/no, then repeat with your own user. Rollback = remove `000-ssh-hardening.conf` and reload.
3. Firewall: the apply enables the inbound Application Firewall (global state on) before stealth, this affects inbound connections on a fresh machine (dresden is already on). Flagged for your sign-off.
4. Casks: LuLu and OverSight install on apply and need manual first-run approval (LuLu is a network system extension; OverSight needs mic/camera permission).
5. SIP is disabled on dresden (the posture reminder surfaces it); re-enabling SIP is Recovery-only and is never automated.

## Effect of merging

Merging changes main only; dresden applies these at D1 (or when you apply interactively, since this slice needs a physically-present console). Nothing here reloads sshd or changes live posture on its own. `dresden`'s live drop-in is still the old one-key `50-` version (the password hole is live) until you apply; apply overwrites it in place and closes the hole.

### DIFF
--- a/.chezmoidata/macos_system_setup.yaml
+++ b/.chezmoidata/macos_system_setup.yaml
@@ -32,6 +32,9 @@ macos:
     - description: "SSH: write the public-key-only sshd drop-in (script self-escalates; reload is a separate operator step)"
       command: '"$HOME/.local/bin/ssh-hardening.sh"'
       sudo: false
+    - description: "Firewall: enable stealth mode (drop unsolicited probes silently)"
+      command: "/usr/libexec/ApplicationFirewall/socketfilterfw --setstealthmode on"
+      sudo: true
   tailnet_pins:
     - fqdn: mister.tail2f2430.ts.net
       ip: "100.109.58.54"
--- a/.chezmoidata/macos_system_setup.yaml
+++ b/.chezmoidata/macos_system_setup.yaml
@@ -32,6 +32,9 @@ macos:
     - description: "SSH: write the public-key-only sshd drop-in (script self-escalates; reload is a separate operator step)"
       command: '"$HOME/.local/bin/ssh-hardening.sh"'
       sudo: false
+    - description: "Firewall: enable the application firewall (required for stealth mode to be active protection)"
+      command: "/usr/libexec/ApplicationFirewall/socketfilterfw --setglobalstate on"
+      sudo: true
     - description: "Firewall: enable stealth mode (drop unsolicited probes silently)"
       command: "/usr/libexec/ApplicationFirewall/socketfilterfw --setstealthmode on"
       sudo: true
--- a/.chezmoidata/macos_system_setup.yaml
+++ b/.chezmoidata/macos_system_setup.yaml
@@ -29,6 +29,9 @@ macos:
     - description: "Install self-healing nix-installer repair LaunchDaemon (NixOS fork)"
       command: '"$HOME/.local/bin/install-nix-repair-hook.sh"'
       sudo: true
+    - description: "SSH: write the public-key-only sshd drop-in (script self-escalates; reload is a separate operator step)"
+      command: '"$HOME/.local/bin/ssh-hardening.sh"'
+      sudo: false
   tailnet_pins:
     - fqdn: mister.tail2f2430.ts.net
       ip: "100.109.58.54"
--- a/.chezmoidata/system_packages_autoinstall.yaml
+++ b/.chezmoidata/system_packages_autoinstall.yaml
@@ -181,6 +181,7 @@ packages:
         - google-chrome
         - karabiner-elements
         - keepassxc
+        - lulu
         - lunar
         - mediosz/tap/swipeaerospace
         - microsoft-auto-update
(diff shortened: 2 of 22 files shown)


webdavis:
Closing as superseded. Everything this branch proposed is on main, delivered through the S9 and S10 slice program rather than as one large PR.

On main today: `dot_local/libexec/macos-defaults/executable_macos-defaults-apply.sh` and `executable_macos-defaults-capture.sh` with their data files under `.chezmoidata/`, and `dot_local/bin/executable_ssh-hardening.sh` with its watchdog and rollback path. Both are documented in CLAUDE.md, and the SSH tooling's reload and lockout-recovery procedure lives in `docs/runbooks/macos-fresh-machine-quickstart.md`.

Reopening this would reintroduce the pre-slice versions, so it is closed rather than merged.

PR #56, opened 2026-07-16
Relocate the osquery feature-set to ~/.local/libexec/osquery/

webdavis:
## Context

The osquery security scripts have lived in `~/.local/bin/`, mixed in with unrelated user tools (trash, herdr, relay, and friends). They are executed by launchd, not typed at a prompt, so they belong in a dedicated `libexec` directory, the conventional home for programs run by other programs. With the directory providing the namespace, the `osquery-` filename prefix becomes redundant and is dropped.

This is the first re-land slice after #54: it moves the five pre-existing scripts so the rest of the osquery feature-set can land on top of the new layout.

## Summary

- Moves the five osquery scripts to `~/.local/libexec/osquery/`, dropping the `osquery-` name prefix.
- Repoints the three launchd plists, the dispatch-library `source` lines, and the enricher path.
- Adds a guard test asserting the new layout: the five scripts exist in libexec and every live reference points there.
- The five scripts now follow the repo's bash style: double-bracket tests, quoted substitutions, self-documenting names.
- Plist filenames and launchd labels are unchanged; dated docs keep old paths as history.

## Notes

- The guard test fails closed: a git error inside its checks fails the test instead of reading as "clean".
- Variable renames are internal only: the cross-script function and environment names are unchanged, and the enricher variable rename is reflected in the guard test.
- Behavior is unchanged; only paths move. Machines applying `main` re-bootstrap the launchd agents via the existing loaders, since the plist content changed.
- The old copies on the live machine are removed as a one-time step at the D1 cutover; that cleanup is deliberately not part of this change.

## Effect of merging

`main` serves the scripts from `~/.local/libexec/osquery/`. No live-machine impact today; dresden applies from `integration/modernization` until the D1 cutover.


dot_local/libexec/osquery/executable_alert-dispatch.sh
webdavis: Always use `[[` instead of `[`

webdavis: Quote sub-processes: `"$(subprocess)"`

webdavis (commented):
All of these files need to be reviewed for following good Bash etiquette:

- [[ instead of [
- Surrounding subprocesses in quotes
- No abbreviated namespaces: variables, function names, etc. Code naming should be clear and readable.

PR #57, opened 2026-07-16
Flag BSD-first stat fallback chains in the test guard

webdavis:
## The problem

Shell scripts ask the `stat` command for file information (permissions, size, and so on). macOS and Linux spell that request differently: Linux (GNU coreutils) uses `stat -c`, macOS (BSD) uses `stat -f`. Portable scripts try one spelling and fall back to the other.

The order of that fallback is everything. GNU-first is safe: on macOS, `stat -c` fails cleanly and the BSD fallback fires. BSD-first is a trap: on Linux, `stat -f` does NOT fail. It is a real GNU option that answers a different question (filesystem status), so it "succeeds" with output that has nothing to do with what the script asked. The fallback never runs and the script silently uses a nonsense value. Like a vending machine that takes the wrong coin and dispenses the wrong snack instead of rejecting it: nothing errors, you just quietly get the wrong thing. This exact mistake broke CI (continuous integration) twice in this repo.

## What this PR does

Two things, both in the test tooling.

First, the test checker learns the stat-order rule. `test/validate-tests.sh` already runs before every test batch to check file placement and permissions; it now also fails loudly on any test file containing a BSD-first `stat` fallback chain, naming the offending file and line. A dedicated test proves the rule both ways: bad chains rejected, correct ones left alone.

Second, the test system was reorganized for clarity. Everything test-related now lives under `test/`, one runner runs any suite, the tests of the test system have their own suite whose directory name says what it is, and every script and function is named for what it does.

This is one slice of re-landing the reverted S9 work in small reviewable pieces.

## The new layout

```
test/
  validate-tests.sh      the checker: placement, permissions, symlinks, stat order
  run-test-suite.sh      the one runner: any suite, optional shuffle and timing
  unit/                  single-component tests, the fast commit gate
  integration/           multi-component tests with stubbed boundaries
  e2e/                   whole-script flows and timing-bound tests
  test-system/           tests of the checker and the runner themselves
    helpers/             sourced function libraries, used only by this suite
  fixtures/              data only, never executed
```

Gone: the `scripts/`-based test tooling, the separate unit-test runner (folded into `run-test-suite.sh`), and the internal word "camp" (everything says "suite" now, in code and prose alike).

## The details

The checker:

- Recognizes four suites: `unit`, `integration`, `e2e`, `test-system`.
- A suite's `helpers/` may hold only non-executable sourced `.sh` files; an executable file or a `.bats` there fails with an explanation.
- Only `validate-tests.sh` and `run-test-suite.sh` may sit at `test/` root; any other stray script fails.
- Fails closed: a `find`, `sort`, `grep`, or `awk` error fails the check rather than passing silently.
- The stat rule recognizes every GNU spelling (`-c`, `--format`, `--printf`), tolerates tabs, extra spaces, and backslash continuations, and analyzes each chain segment separately so an unrelated earlier GNU call cannot mask a later bad chain on the same line.

The runner:

- One script runs any suite; `--shuffle` (seeded, replayable) and `--warn-slow-ms` (warn-only slow-test summary) are options the unit suite uses.
- Option values must be plain unsigned integers, checked before any test runs; `TEST_SEED` and `UNIT_WARN_MS` environment variables still work.
- The shuffled list is verified against discovery; a shuffler that drops or adds entries fails the run instead of silently skipping tests.
- Child tests inherit the caller's locale and run with file descriptor 3 closed, so a test that reads stray input cannot swallow the test queue.

Structure:

- Each suite's bats files run exactly once per `just test`; the old aggregate re-run is gone.
- The test-system tests collect every failed assertion and report them all at the end of a run.
- Executed scripts are functions ending in `main "$@"`; helpers are sourced function files with no main.

## What it does not catch

- The guarantee covers literal chains in raw text only. Runtime-assembled chains (eval over a variable holding the BSD form, strings handed to `sh -c`, here-docs) are out of scope, and passing tests pin that boundary so any future widening shows up as a test change.
- Comments and fixture prose ARE scanned, on purpose: a commented example gets copy-pasted, and a sourced fixture carries the same trap. Inert prose containing the bad pattern trips the gate by design; reword it or assemble the stat tokens at run time.
- The checker lives inside its own scan root, so its comments describe the bad pattern without spelling it literally; `test/test-system/stat-order.sh` is the authoritative documentation, with every rule pinned as a named fixture.

## Effect of merging

This is dev/CI tooling only; there is no runtime or live-machine impact. The `just` recipe names are unchanged except one: `test-guard` is now `validate-tests`, matching the script it runs.


scripts/test-guard.sh
```diff
@@ -66,3 +66,105 @@ if [[ -n $bad ]]; then
   printf 'Fix placement/mode (and REPO_ROOT depth is ../.. inside a camp).\n' >&2
   exit 1
 fi
```
webdavis: Change `f` to `file`. Variable names should not be abbreviations.

scripts/test-guard.sh
```diff
@@ -66,3 +66,105 @@ if [[ -n $bad ]]; then
...
+if [[ $grep_status -gt 1 ]]; then
+  printf 'FAIL: stat-chain candidate scan of %s/ failed (grep exit %d); refusing to pass on a partial scan\n' "$root" "$grep_status" >&2
+  exit 1
+fi
+
+bsd_first_chains=""
+while IFS= read -r scanned_file; do
+  [[ -n $scanned_file ]] || continue
+  if ! LC_ALL=C awk '
+    function flush(   line_copy, segment_count, i, segment, bsd_index, gnu_index) {
```
webdavis: Why is there whitespace between the opening parenthesis `flush` signature?

Additionally, this Awk function is complex and needs either a comment to explain it, or explained through code/name-spacing more clearly. And/or abstractions within the embedded Awk code, if possible.

scripts/test-guard.sh
```diff
@@ -66,3 +66,105 @@ if [[ -n $bad ]]; then
...
+# line 2 held `||` but no `stat -f` (and a GNU-first chain split the same way
+# false-POSITIVED on the `|| stat -f` continuation line). Join backslash
+# continuations into one logical line, keyed by the starting physical line number,
+# before matching.
+# Fail closed: candidate discovery and per-file matching are CHECKED commands
+# into temp files, mirroring the discovery pipeline above. grep exit 1 means
+# "no candidates" (a pass); anything above 1 is a tool error and MUST fail the
+# guard, never silently yield an empty candidate list and a green pass. An awk
+# failure likewise fails the guard instead of vanishing inside a process
+# substitution. No `2>/dev/null` -- a real tool error is seen.
```
webdavis: This comment is very long and either needs more structured framing, but realistically since it's not documented as code, it will eventually drift. Comments are okay, but only if they are carefully documented.  Not sure what to do here — we need to think harder about this important documentation issue.

scripts/test-guard.sh
```diff
@@ -66,3 +66,105 @@ if [[ -n $bad ]]; then
...
+stat_candidates_list="$(mktemp)"
+chain_lines_list="$(mktemp)"
+trap 'rm -f "$links_list" "$files_list" "$stat_candidates_list" "$chain_lines_list"' EXIT
+
+# Token matching is whitespace-tolerant (`stat[[:space:]]+-f`): a tab or a run
+# of spaces between `stat` and its flag is the same command and must not bypass
+# the prefilter or the per-segment matching below.
+# LC_ALL=C pins byte semantics: BSD grep's -I binary/text classification (and
+# regex character classes) vary with the caller's locale, C does not.
```
webdavis: Not sure I fully understand this comment/explanation.

test/tools/test-guard.sh
```diff
@@ -66,3 +66,105 @@ if [[ -n $bad ]]; then
...
+      if (line ~ /\\[ \t]*$/) { sub(/\\[ \t]*$/, " ", line); joined = joined line; next }
+      joined = joined line
+      flush()
+    }
+    END { flush() }
+  ' "$scanned_file" >"$chain_lines_list"; then
+    printf 'FAIL: stat-chain scan of %s failed (awk error); refusing to pass on a partial scan\n' "$scanned_file" >&2
+    exit 1
+  fi
+  while IFS= read -r chain_start_line; do
```
webdavis: Line separation needed between ending if-statement (`fi`) and the start of the while-loop.

scripts/test-guard.sh
webdavis: If `links_list` is for symlinks then it should be named accurately.

test/tools/test-guard.sh
```diff
@@ -0,0 +1,164 @@
+#!/usr/bin/env bash
+# test-guard.sh [root] -- placement / mode / symlink guard for the test pyramid.
```
webdavis: "pyramid" is a confusing word to use here.

test/tools/test-guard.sh
```diff
@@ -0,0 +1,164 @@
...
+# A dependency of every test recipe (root defaults to `test`). It fails when a
+# test file cannot be seen by, or could escape, the gate:
+#
+#   - a *.sh OR *.bats not sitting DIRECTLY in a recognized camp
```
webdavis: We should use a better word than "camp".

test/tools/test-guard.sh
```diff
@@ -0,0 +1,164 @@
...
+#     (fixture data and sourced libs, never run directly);
+#   - a camp *.sh that is not executable (invisible to the runner's -perm probe);
+#   - ANY symlink below test/. A physical `find -type f` skips symlinked files
+#     and symlinked camp dirs, so a tracked symlink would evade this guard and
+#     every gate. Following it risks out-of-tree traversal and cycles, so the
+#     guard REJECTS symlinks rather than resolving them.
+#
+# Discovery is a CHECKED foreground pipeline (pipefail on) into a temp file, not
+# a process substitution: a traversal or sort error must FAIL the guard, never
+# yield a short list and a green pass. No `2>/dev/null` -- a real error is seen.
```
webdavis: This is worded in a confusing way. Simplify it. Don't make it any longer than it already is —shorter, if possible.

test/tools/test-guard.sh
```diff
@@ -0,0 +1,164 @@
...
+set -euo pipefail
+
+root="${1:-test}"
```
webdavis: This needs a short comment explaining why the test-root folder needs to be accepted as an argument.

test/tools/test-guard.sh
```diff
@@ -0,0 +1,164 @@
...
+
+# Symlink rejection first. `-type l` matches symlinked files AND symlinked dirs
+# (find does not follow symlinks without -L), catching a symlinked camp dir too.
+if ! find "$root" -type l -print0 >"$symlink_paths_list"; then
+  printf 'FAIL: symlink scan of %s/ failed\n' "$root" >&2
+  exit 1
+fi
+
+if [[ -s $symlink_paths_list ]]; then
+  printf 'FAIL: symlinks are not allowed below %s/ (out-of-tree traversal / cycle risk); remove:\n' "$root" >&2
```
webdavis: This failure message is confusing and could be worded more clearly.

test/tools/test-guard.sh
```diff
@@ -0,0 +1,164 @@
...
+while IFS= read -r -d '' file; do
+  case "$file" in
+    "$root"/fixtures/*) continue ;;
+    # Test infrastructure (this guard and the camp runners), run by just, never discovered as tests.
+    "$root"/tools/*) continue ;;
+    "$root"/unit/*/* | "$root"/integration/*/* | "$root"/e2e/*/*)
+      bad+="$file (nested; camps are flat)"$'\n'
+      ;;
+    "$root"/unit/*.sh | "$root"/integration/*.sh | "$root"/e2e/*.sh)
+      [[ -x $file ]] || bad+="$file (not executable; invisible to the gate)"$'\n'
```
webdavis: The description in the parenthesis needs to be more clear.

test/tools/test-guard.sh
```diff
@@ -0,0 +1,164 @@
...
+      ;; # bats live flat in a camp; bats itself runs them (no +x needed)
+    *)
+      bad+="$file (outside the unit/integration/e2e camps)"$'\n'
+      ;;
+  esac
+done <"$files_list"
+
+if [[ -n $bad ]]; then
+  printf 'FAIL: misplaced or misconfigured test scripts:\n%s' "$bad" >&2
+  printf 'Fix placement/mode (and REPO_ROOT depth is ../.. inside a camp).\n' >&2
```
webdavis: The fix instruction is confusing and hard to understand for the reader.

test/tools/test-guard.sh
```diff
@@ -0,0 +1,164 @@
...
+#     fixture prose count; the scan reads raw text on purpose (copy-paste risk).
+#   - ALLOWED: GNU-first chains, and capability-gated bare BSD-form calls.
+#   - Boundary: literal chains in raw text only. Runtime-assembled chains and a
+#     masking GNU call inside the same unseparated segment are out of scope.
+#   - Fail closed: a grep or awk error fails the guard, never a silent pass.
+#
+# Each rule and boundary here is pinned as a named fixture + assertion in
+# test/integration/test-guard-bsd-stat.sh; that test is the authoritative
+# documentation and cannot drift. (The guard lives inside its own scan root, so
+# no comment here may spell a literal BSD-first chain; hence this phrasing.)
```
webdavis: This entire comment is confusing and difficult to understand. Simplify and shorten it. If it's not relevant for a first time reader to understand, then it should be removed.

test/tools/test-guard.sh
```diff
@@ -0,0 +1,164 @@
...
+stat_candidates_list="$(mktemp)"
+chain_lines_list="$(mktemp)"
+trap 'rm -f "$symlink_paths_list" "$files_list" "$stat_candidates_list" "$chain_lines_list"' EXIT
+
+# Candidate discovery, checked: grep exit 1 means "no candidates" (a pass);
+# anything above 1 is a tool error and fails the guard. Token matching is
+# whitespace-tolerant (a tab or a run of spaces between `stat` and its flag is
+# the same command). LC_ALL=C because grep decides whether a file is binary
+# differently depending on the machine's locale; pinning the locale makes every
+# machine decide identically.
```
webdavis: This comment is confusing and difficult to understand. Comments should be simply and easy for developers to understand. They shouldn't be missing context either.

test/integration/test-guard-bsd-stat.sh
```diff
@@ -0,0 +1,274 @@
...
+# and must stay allowed. This drives the guard against a scratch fixture tree.
+#
+# Self-immunity trick: the two stat tokens are assembled from the variables
+# below, never written as a literal BSD-first chain in THIS file. `just test`
+# runs the real guard over test/, which scans this very file; keeping every line
+# that also carries `||` off the literal tokens lets the fixtures stay honest
+# without the guard flagging its own test.
+set -euo pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
```
webdavis: The repo root should be found using `git`, so that this is more resilient to changes.

test/integration/test-guard-bsd-stat.sh
```diff
@@ -0,0 +1,274 @@
...
+  printf 'test-guard-bsd-stat: FAIL -- guard missing or not executable: %s\n' "$GUARD" >&2
+  exit 1
+}
+
+# The GNU form and the BSD form, as tokens (neither assignment carries `||`).
+gnu_form='stat -c'
+bsd_form='stat -f'
+
+# Whitespace variants (a tab, multiple spaces): legal shell spellings of the
+# same commands, derived from the tokens so no literal appears here either.
```
webdavis: This comment is hard to understand and needs to be reworded.

test/integration/test-guard-bsd-stat.sh
```diff
@@ -0,0 +1,274 @@
...
+bsd_form_tab="${bsd_form/ /$'\t'}"
+gnu_form_wide="${gnu_form/ /   }"
+
+# GNU long-option spellings. These literals are safe to write here: only a BSD
+# form inside a `||` segment can trip the guard, never a GNU form.
+gnu_long_form='stat --format'
+gnu_printf_form='stat --printf'
```
webdavis: I don't understand this.

test/integration/test-guard-runner.sh
webdavis: This comment is confusing and needs to be improved.

webdavis (commented):
For many of these files, logic could be moved into functions so that behavior is contained and modular, instead of linear.

test/test-system/stat-order.sh
```diff
@@ -0,0 +1,289 @@
...
+
+  # (n) A BSD-first chain inside \$root/fixtures/ -- MUST be flagged: the
+  # placement check exempts fixtures/, but the stat scan reads every text file
+  # below the scanned root (a sourced fixture lib carries the same trap).
+  local fixtures_lib="$flagged_root/test/fixtures/stat-lib.sh"
+  mkdir -p "$flagged_root/test/fixtures"
+  {
+    printf '#!/usr/bin/env bash\n'
+    printf '%s\n' "perms() { $bsd_form '%Lp' \"\$1\" || $gnu_form '%a' \"\$1\"; }"
+  } >"$fixtures_lib"
```
webdavis: Should the fixture captures be moved to a function for modularlity?

test/test-system/helpers/capture-output.sh
```diff
@@ -0,0 +1,18 @@
...
+captured_output=""
+# shellcheck disable=SC2034 # read by the scripts that source this helper
+captured_status=0
+
+capture_output() {
+  set +e
+  captured_output="$("$@" 2>&1)"
+  captured_status=$?
+  set -e
+}
```
webdavis, line 37: Helper scripts shouldn't introduce global state. They should use Nameref output params, instead.

test/test-system/placement.sh
```diff
@@ -0,0 +1,197 @@
...
+run_guard() {
+  local root="$1"
+  shift
+  capture_output env "$@" "$GUARD" "$root"
+}
+
+# Set in main; global so the EXIT trap can still see it after main returns.
+work=""
+
+main() {
```
webdavis: A lot of the logic in `main` could be moved to independent functions within this script, where `main` calls each function.

test/test-system/runner-discovery.sh
```diff
@@ -0,0 +1,128 @@
...
+run_suite() {
+  local suite="$1"
+  shift
+  capture_output env "$@" "$RUN_SUITE" "$suite"
+}
+
+# Set in main; global so the EXIT trap can still see it after main returns.
+work=""
+
+main() {
```
webdavis: A lot of the logic in main could be moved to independent functions within this script, where main calls each function.

test/run-test-suite.sh
```diff
@@ -0,0 +1,252 @@
...
+  # bats (the usual case -- `just test` runs on the host), fall back into the
+  # Nix devshell, mirroring the aggregate `test` recipe.
+  if command -v bats >/dev/null 2>&1; then
+    bats --jobs 4 "${bats_files[@]}" || status=1
+  else
+    nix develop .#run --command bats --jobs 4 "${bats_files[@]}" || status=1
+  fi
+}
+
+main() {
```
webdavis: A lot of the logic in main could be moved to independent functions within this script, where main calls each function.

test/validate-tests.sh
```diff
@@ -0,0 +1,203 @@
...
+main() {
+  local root="${1:-test}"
+  workdir="$(mktemp -d)"
+  trap 'rm -rf "$workdir"' EXIT
+  check_symlinks "$root" "$workdir" || exit 1
+  check_placement "$root" "$workdir" || exit 1
+  check_stat_order "$root" "$workdir" || exit 1
+}
+
+main "$@"
```
webdavis, line 218: This is a great example of how `main` should look, making calls to other functions to execute logic. This reads well and a developer can easily look at this and know what to do. It's short and concise.

webdavis (commented):
I left a comment regarding `main` in `test/validate-tests.sh`. That's how `main` should look in a script: short, concise, and allowing functions to do the work.

test/test-system/helper-functions.sh
```diff
@@ -0,0 +1,83 @@
...
+source "$here/helpers/capture-output.sh"
+# shellcheck source=helpers/create-test-trees.sh
+source "$here/helpers/create-test-trees.sh"
+# shellcheck source=helpers/report-test-failures.sh
+source "$here/helpers/report-test-failures.sh"
+
+# Set in main; global so the EXIT trap can still see it after main returns.
+work=""
+
+main() {
```
webdavis: `main` is for control flow. Move the logic in main into functions.

test/validate-tests.sh
```diff
@@ -0,0 +1,208 @@
...
+    return 1
+  fi
+
+  local bsd_first_chains="" scanned_file chain_start_line
+  while IFS= read -r scanned_file; do
+    [[ -n $scanned_file ]] || continue
+    # awk idiom: awk has no `local`; extra function parameters ARE the locals,
+    # and the wide gap in flush()'s parameter list separates real arguments
+    # (none here) from those locals.
+    if ! LC_ALL=C awk '
```
webdavis: Move this awk implementation to an isolated function, called by `check_stat_order`

webdavis (commented):
Fix these changes. Review agent should be isolated to the files changed and collaborators so that review doesn't take too long. Once these changes are fixed and tests pass we can merge this.

PR #58, opened 2026-07-19
Re-land the osquery dispatch library with save-first alert durability

webdavis:
## What this PR does

This PR re-lands the upgrade to the osquery alert dispatch library, the shared helper that
every osquery producer uses to raise an alert. It also adds the repository's first
cross-suite test helpers and the two behavior tests that exercise the library.

It is one piece of re-landing the previously reverted S9 work in small, reviewable steps.

## The reliability change

The headline is a durability fix for how a critical alert reaches its destination. Think of
mailing an important letter: the old approach dropped the letter straight into the slot and
only kept a copy if the post office refused it. If the mail carrier stumbled between the slot
and the truck, the letter was simply gone, and a lost alert looks exactly like "nothing is
wrong."

The dispatcher now writes a copy of every alert to disk BEFORE it attempts to send, and
deletes that copy only after the send is confirmed. A crash or a kill at any point during the
send can no longer lose the alert: the saved copy remains and is retried on the next run, in
the order the events actually happened. This is stronger than the original behavior, which
saved a copy only after a send had already failed.

## Summary

- Undelivered alerts are saved to disk before the send and deleted only on a confirmed success.
- A partially written record left by a crash is quarantined for inspection, never silently skipped.
- Retries deliver the oldest alert first, ordered by the time each event was recorded.
- Only CRIT pages route to the priority channel, and each page is marked with its tier.
- The webhook signing key never appears on a command line; it is read from its 0600 file and used without argv exposure.
- A new `test/helpers/` directory holds the first test helpers shared across suites.

## The details

- A dedicated undelivered-alerts directory holds one mode-600 record per alert in a mode-700
  directory, and a retry function replays them to the local webhook only.
- Each record is written atomically: a temp file is written, flushed, then renamed into place,
  so a reader never sees a torn entry. The encoded body and timestamp are validated before the
  record is opened, so an encoding failure fails the save instead of storing an empty record.
- The signature is computed by a small HMAC-SHA256 signer that takes the key in process memory
  and hashes only the message bytes, so the key is never passed as a command-line argument. Its
  output is pinned byte-identical to the reference `openssl dgst -hmac` on both BSD (LibreSSL)
  and GNU (OpenSSL) toolchains.
- The shared helpers return their results through nameref output parameters and reject colliding
  or reserved destination names, so a caller cannot accidentally alias its own variables.
- The placement checker now permits `test/helpers/` under the same rule as a suite's helpers:
  non-executable sourced `.sh` only, with executable files and test files there rejected.
- Two tests cover the library: an integration test for severity routing and body shape, and an
  end-to-end test for the save, retry, ordering, and key-safety behaviors.

## Effect of merging

The dispatch library and its tests land under the libexec home. The periodic retry drainer that
calls the retry function on a schedule arrives with the alerter in a later slice, so this change
is inert on its own beyond the library and its tests. There is no live-machine impact: dresden
applies from `integration/modernization` until D1.

### DIFF
--- a/dot_local/libexec/osquery/executable_alert-dispatch.sh
+++ b/dot_local/libexec/osquery/executable_alert-dispatch.sh
@@ -1,22 +1,21 @@
 #!/usr/bin/env bash
 #
 # alert-dispatch.sh, sourced helper, not run directly. Provides
-# send_alert(), which dispatches one finding to BOTH the local macOS notifier
-# (alerter) and a hermes Discord webhook, routing by severity (CRIT -> #priority,
-# else -> #osquery). Signing, dual-channel delivery, and durable handling of an
-# undelivered page all live here so the three producers (results-alerter.sh,
-# firewall-gatekeeper-monitor.sh, and uptime-watchdog.sh) share one
-# implementation.
+# send_alert(), which always fires the local macOS notifier (alerter) and, for a
+# CRIT severity ONLY, POSTs the page to the hermes #priority Discord webhook. v2
+# has NO #osquery channel: there is deliberately no non-priority route for a
+# producer to leak to. Signing and durable handling of an undelivered page live
+# here so the three producers (results-alerter.sh, firewall-gatekeeper-monitor.sh,
+# and uptime-watchdog.sh) share one implementation.
 #
 # Usage (from a sourcing script):
 #   source "$HOME/.local/libexec/osquery/alert-dispatch.sh"
 #   send_alert CRIT "Firewall disabled" "alf global_state 1 -> 0" Sosumi
 
-# Two routes, same app (osquery), each signed with the one osquery key below.
-# CRIT findings go to the #priority channel (the one channel the user watches);
-# everything else goes to the quiet #osquery log channel. send_alert picks the
-# URL from its severity argument.
-OSQUERY_HERMES_URL="${OSQUERY_HERMES_URL:-http://127.0.0.1:8644/webhooks/osquery}"
+# One Discord route: the #priority channel (the one channel the user watches),
+# signed with the osquery HMAC key below. v2 has NO #osquery channel, only a
+# confirmed CRIT page is POSTed; any other severity does the local notification
+# only. There is deliberately no non-priority URL for a producer to leak to.
 OSQUERY_HERMES_PRIORITY_URL="${OSQUERY_HERMES_PRIORITY_URL:-http://127.0.0.1:8644/webhooks/osquery-priority}"
 # The notifier signs with its OWN copy of the HMAC key, read from its own secret
 # file, NOT from hermes's .env. HMAC is symmetric so the value must match the
@@ -138,10 +137,6 @@ retry_undelivered_alerts() {
 send_alert() {
   local severity="$1" title="$2" detail="$3" sound="${4-}" occurrence="${5-}"
 
-  # Route by severity: only CRIT reaches #priority.
-  local webhook_url="$OSQUERY_HERMES_URL"
-  [[ $severity == "CRIT" ]] && webhook_url="$OSQUERY_HERMES_PRIORITY_URL"
-
   # The local notifier renders plain text, so strip Discord markdown (**bold**,
   # `code`) for it; the webhook POST below keeps the markdown intact.
   local plain_title plain_detail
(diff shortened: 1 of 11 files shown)


dot_local/libexec/osquery/executable_alert-dispatch.sh
```diff
@@ -1,47 +1,246 @@
...
+# has NO #osquery channel: there is deliberately no non-priority route for a
+# producer to leak to. Signing and durable handling of an undelivered page live
+# here so the three producers (results-alerter.sh, firewall-gatekeeper-monitor.sh,
 # and uptime-watchdog.sh) share one implementation.
 #
+# The undelivered-alerts store is WRITE-AHEAD: send_alert persists the page to
+# disk BEFORE the first network attempt and deletes it only after a confirmed 2xx.
+# A crash or kill anywhere between persist and success therefore leaves a
+# recoverable record for the next drain (a lost page is indistinguishable from
+# "all clear").
```
webdavis: What does `a lost page is indistinguishable from "all clear"` mean? The comment length should be roughly the same.

dot_local/libexec/osquery/executable_alert-dispatch.sh
```diff
@@ -1,47 +1,246 @@
...
 # alert-dispatch.sh, sourced helper, not run directly. Provides
-# send_alert(), which dispatches one finding to BOTH the local macOS notifier
-# (alerter) and a hermes Discord webhook, routing by severity (CRIT -> #priority,
-# else -> #osquery). Signing and dual-channel delivery live here so the three
-# producers (results-alerter.sh, firewall-gatekeeper-monitor.sh,
+# send_alert(), which always fires the local macOS notifier (alerter) and, for a
+# CRIT severity ONLY, POSTs the page to the hermes #priority Discord webhook. v2
+# has NO #osquery channel: there is deliberately no non-priority route for a
+# producer to leak to. Signing and durable handling of an undelivered page live
+# here so the three producers (results-alerter.sh, firewall-gatekeeper-monitor.sh,
```
webdavis: Don't mention things that don't exist. For example, the `#osquery` Discord channel. If a feature or object doesn't exist or has been deleted then it doesn't need to live on through comments. This is explicit drift.

dot_local/libexec/osquery/executable_alert-dispatch.sh
```diff
@@ -1,47 +1,246 @@
...
 
-# Two routes, same app (osquery), each signed with the one osquery key below.
-# CRIT findings go to the #priority channel (the one channel the user watches);
-# everything else goes to the quiet #osquery log channel. send_alert picks the
-# URL from its severity argument.
-OSQUERY_HERMES_URL="${OSQUERY_HERMES_URL:-http://127.0.0.1:8644/webhooks/osquery}"
+# One Discord route: the #priority channel (the one channel the user watches),
+# signed with the osquery HMAC key below. v2 has NO #osquery channel, only a
+# confirmed CRIT page is POSTed; any other severity does the local notification
+# only. There is deliberately no non-priority URL for a producer to leak to.
```
webdavis: Again, don't mention features that don't exist.

dot_local/libexec/osquery/executable_alert-dispatch.sh
```diff
@@ -1,47 +1,246 @@
...
 OSQUERY_HERMES_PRIORITY_URL="${OSQUERY_HERMES_PRIORITY_URL:-http://127.0.0.1:8644/webhooks/osquery-priority}"
 # The notifier signs with its OWN copy of the HMAC key, read from its own secret
 # file, NOT from hermes's .env. HMAC is symmetric so the value must match the
 # gateway's, but the signer must not reach into the verifier's credential store;
 # each side owns its own copy. Single-value file, mode 600, runtime (not tracked).
 OSQUERY_WEBHOOK_SECRET_FILE="${OSQUERY_WEBHOOK_SECRET_FILE:-$HOME/.config/osquery/webhook-secret}"
 OSQUERY_DELIVERY_LOG="${OSQUERY_DELIVERY_LOG:-$HOME/.local/log/osquery/webhook-delivery.log}"
+# Undelivered pages are stored here, one mode-600 file per page in a mode-700
+# dir, so a transient gateway outage never loses a page. retry_undelivered_alerts
```
webdavis: dir → directory. We try not to abbreviate words.

dot_local/libexec/osquery/executable_alert-dispatch.sh
```diff
@@ -1,47 +1,246 @@
...
+# replays them in occurrence-time order.
+OSQUERY_UNDELIVERED_ALERTS_DIR="${OSQUERY_UNDELIVERED_ALERTS_DIR:-$HOME/.local/state/osquery-undelivered-alerts}"
+
+# A monotonic per-process sequence so the fallback request id (when no occurrence
+# identity is threaded) is unique across calls in one process.
+_OSQUERY_ALERT_SEQUENCE=0
 
 # Append a timestamped line to the delivery log (best-effort; never fails caller).
+# Only metadata is ever logged, never the body or the HMAC secret.
```
webdavis: Be more clear than "best-effort; never fails caller".

dot_local/libexec/osquery/executable_alert-dispatch.sh
```diff
@@ -1,47 +1,246 @@
...
+  # shellcheck disable=SC2059 # the format is a fixed \xHH byte string, no user data
+  inner_hex="$({
+    printf "$ipad_format"
+    cat
+  } | openssl dgst -sha256 | awk '{print $NF}')"
+  for ((i = 0; i < ${#inner_hex}; i += 2)); do inner_format+="\\x${inner_hex:i:2}"; done
+  # outer = SHA256( (K' xor opad) || inner ).
+  # shellcheck disable=SC2059 # the format is a fixed \xHH byte string, no user data
+  printf "$opad_format$inner_format" | openssl dgst -sha256 | awk '{print $NF}'
+}
```
webdavis, line 89: Since `openssl dgst -sha256` is sensitive, in that bytes must be hashed only on STDIN, it should be protected by its own function call to isolate and protect that behavior. We can think about this as a security protocol, and a protocol in the more general software engineering sense.

dot_local/libexec/osquery/executable_alert-dispatch.sh
```diff
@@ -66,42 +265,93 @@ send_alert() {
...
+  signature="$(printf '%s' "$body" | _hmac_sha256_hex "$secret")"
 
   for attempt in 1 2 3; do
     http_status="$(curl -s -o /dev/null -w '%{http_code}' --max-time 5 \
-      -X POST "$webhook_url" \
+      -X POST "$url" \
       -H 'Content-Type: application/json' \
       -H "X-Webhook-Signature: $signature" \
       -H "X-Request-ID: $request_id" \
       --data "$body")" || http_status=000
```
webdavis: This `curl` command is complex enough that it should be wrapped in its own isolated function, and then invoking functions will call that function. Theoretically, this also protects invoking code from query refactoring's and about provides some level of abstraction in Bash.

dot_local/libexec/osquery/executable_alert-dispatch.sh
```diff
@@ -66,42 +265,93 @@ send_alert() {
     fi
   fi
 
+  # v2: only a CRIT page is delivered to Discord. Any other severity stops here,
+  # after the local notification, there is no #osquery channel to POST to.
```
webdavis: Mentions `#osquery` again.

dot_local/libexec/osquery/executable_alert-dispatch.sh
```diff
@@ -1,47 +1,246 @@
...
+# (distinct incidents survive) yet stable across a retry of the same occurrence
+# (the gateway dedups it). Absent means a per-call unique id.
+#
+# Delivery is WRITE-AHEAD: the page is persisted BEFORE the first network attempt
+# and deleted only after a confirmed 2xx. Return contract: 0 when the page was
+# DELIVERED or durably STORED; NONZERO only on a HARD failure (the write-ahead
+# persist itself failed, so the page is neither delivered nor stored), after
+# firing a loud local alert, so the caller does not advance its cursor past a page
+# that was lost.
+send_alert() {
```
webdavis, line 359: This function is huge and packed with behavior. Much of this behavior can be moved to helper functions that should then invoked by `send_alert`.

webdavis (commented):
Take care of these and then this PR should be good to merge.

Additionally, use a naming system for namerefs. If the variable is a nameref inside a function, then the function-scoped variable should start with the function name. For example:

```bash
example_function_name() {
  # It's okay if nameref variables are longer. This practice prevents namespace collisions.
  local -n example_function_name_<variable_name>=$<#>
}
```

PR #94, opened 2026-07-28
fix(ssh): read every configuration line the way sshd reads it

webdavis:
## Context

The tool that checks this machine's remote-login policy has to model how the login daemon reads its own
configuration file, so it can spot a rule that quietly re-opens password access for some visitors but not
the one it happens to be testing from. Wherever that model disagrees with the daemon, a file can be read
as a real rule by the daemon and as nothing at all by the checker.

That has now happened five times, each found by a separate review and patched one at a time. This closes
the outstanding ones and, more usefully, replaces the guess-and-patch loop with a test that compares the
model against the daemon itself over many inputs.

## Summary

- Closes a hole where a line beginning with a discarded character was invisible to the checker.
- Closes a second hole in how quoted file paths were read, which decides which files get examined at all.
- Adds a test that asks the real daemon what each of sixty-eight lines means and requires the checker to
  agree.
- The test found three families of disagreement on its first run, not the one it was written for.
- Says plainly what the test does and does not cover, in place of a completeness claim that has been
  wrong four times.

Key files: `dot_local/bin/executable_ssh-hardening.sh`, one new test.

## What was actually wrong

Two distinct defects, both confirmed against the real daemon before anything was written.

The first: a line whose opening character is discarded. The daemon throws that empty piece away and
treats the next word as the setting name, so such a line is a live rule. The checker saw an empty name,
matched nothing, and moved on. Twenty-one variations of this, including forms that hide an entire
conditional block.

The second was not on anyone's list. A file can pull in other files, and the path it names is parsed by a
different set of rules than a setting name is. Quoted paths, escaped spaces, and a path containing a
control character are all accepted by the daemon and really do pull the file in, and none of them were
followed. Every earlier review had treated setting names as the surface.

Both are the same class as the three found before them, and the previous fixes did not prevent either.

## Why a test rather than a fifth patch

The rules were re-derived by experiment against the installed daemon rather than from its manual or from
the comment block, which had already been wrong twice. Six rounds of probes established what it accepts,
what it rejects, and what each accepted form resolves to for a visitor who is not this machine.

Those probes became the test. For each of sixty-eight lines it asks the daemon whether the file is
acceptable, then what an outside visitor would get, and then requires the checker to refuse exactly the
dangerous ones and stay quiet on the harmless ones. Run against the unfixed tool, twenty-seven of the
sixty-eight slipped through, in three separate families. Run against the fixed one, none do, and none of
the harmless forms raises a false alarm either.

The value is not the sixty-eight lines. It is that a disagreement nobody has thought of now fails a test
instead of waiting for the next review to stumble on it, including one introduced by a future release of
the daemon itself.

## What is deliberately still open

Three test mutations survive, and each is reported rather than hidden. One concerns a path style the test
harness cannot reproduce faithfully without privileges it should not have. Two are guards that are inert
against today's code but make a stated contract true rather than accidentally true, and were kept for
that reason.

The corpus is bounded, not exhaustive, and both the code comment and the test header say so. That
replaces the completeness claim that stood there before, which was wrong every time it was written.

## How this relates to the change merged alongside it

The apply-safety work merged separately and independently fixed one of the same five families. This
branch carries the more faithful version of that shared area, kept after running both side by side on
fifteen probe forms: they agree everywhere both handle, and one continues to agree where the other stops.
Every guard the other change introduced was re-tested here and still holds.

## Effect of merging

Advances `main` only. The live machine follows another branch, and this changes only what the checker
reads, never what the daemon is configured to do.

### DIFF
--- a/dot_local/bin/executable_ssh-hardening.sh
+++ b/dot_local/bin/executable_ssh-hardening.sh
@@ -393,8 +393,12 @@ check_global() {
 # line resolves to -- and not from sshd_config(5), which documents almost none
 # of it.
 #
-#   THE LINE. Leading whitespace is stripped before anything else, and a line
-#   that is then empty or opens with '#' is dropped.
+#   THE LINE. Trailing space, tab, carriage return and FORM FEED are trimmed
+#   off the whole line before either tokenizer runs; vertical tab, BEL and BS
+#   are NOT in the set (all seven measured). A CRLF line ending -- every line
+#   of a file with Windows line endings -- is the everyday case this trim
+#   exists for. Leading whitespace is stripped next, and a line that is then
+#   empty or opens with '#' is dropped.
 #
 #   THE KEYWORD (strdelim semantics). Space, tab and CARRIAGE RETURN separate;
 #   a single '=' may stand in for the separating whitespace and is consumed
@@ -435,6 +439,7 @@ check_global() {
 
 CONFIG_TAB=$'\t'
 CONFIG_CR=$'\r'
+CONFIG_FF=$'\f'
 # Written in ANSI-C quoting like the two constants above, and not as a plain
 # quoted backslash: shellcheck reads '\' as a botched attempt to escape a
 # single quote (SC1003) and refuses it, while shfmt -s rewrites "\\" into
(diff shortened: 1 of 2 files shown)


webdavis:
**Do not merge.** An independent review found that this branch makes one input class worse than `main`,
in the exact area it exists to fix. Reproduced independently before posting.

A configuration line that pulls in another file, terminated with a carriage return before the newline,
which is what any file with Windows line endings looks like:

```
Match Address *,!127.0.0.1
Include <path>/payload.conf<CR>
```

The daemon strips a trailing carriage return from the end of a line, so it really does pull that file in
and really does re-open password access for every visitor that is not this machine. Measured:

```
real daemon, off-machine address   ->  password authentication: yes
this branch                        ->  the check PASSES   (exit 0)
main                               ->  the check FAILS    (exit 1, correct)
```

Root cause: the daemon trims trailing space, tab, carriage return and form feed from the whole line. This
branch correctly stopped treating a carriage return as a separator inside a path, which is right, but did
not add the line-level trim, so every carriage-return-terminated line now carries that byte into the
filename it looks for. No file matches, the pulled-in file is never examined, and the daemon reads it
anyway.

The same cause breaks the other direction too, and that half is arguably worse: a file with Windows line
endings that merely restates a correct setting now FAILS the check, with a message that reads as
nonsense because the offending byte is invisible:

```
sets 'passwordauthentication no' inside a Match block (want 'no')
```

Install uses this check as its gate, so such a file would make the tool roll back its own hardening and
refuse to install.

A second, smaller regression: a backslash followed by a tab is a separator to the daemon, not an escape,
so a path containing one is merged into a neighbouring argument and the pulled-in file is missed.

Neither is a pre-existing problem. Both arrived with this branch. The pre-existing backslash-and-glob
class the review also found is real but predates this work and belongs in a follow-up.

The rest of the branch holds up: eight of eight mutations against the corpus were caught, the twelve
rejected forms fail through the intended path, and the inert forms are genuinely asserted rather than
merely recorded. The corpus is doing its job. It simply does not yet contain the forms above, which is
why the regression got through.

Fix in progress on the same branch. This comment stays until the reproduction above returns a failing
check on both counts.

webdavis:
Both regressions are fixed and verified independently at `a7b7ded9`.

The carriage-return case now behaves correctly in both directions:

```
Include <path>/payload.conf<CR>        check FAILS (was passing)
hardened restatement with CRLF         check PASSES (was falsely failing)
the two original hostile fixtures      still refused
```

Root cause closed by trimming trailing whitespace from the whole line before tokenizing, matching the daemon. The trim set was measured rather than assumed: space, tab, carriage return and form feed are trimmed; vertical tab, BEL and backspace are not. That boundary is now pinned from both sides, so a future edit that trims too much fails as loudly as one that trims too little.

The backslash-then-tab case is fixed too. The daemon treats that pair as a separator, not an escape, which is asymmetric with backslash-space. Both the code and the comment that stated it wrongly are corrected.

Most importantly, the branch is now demonstrably at least as strict as `main` on every one of the 76 forms:

| | |
|---|---|
| forms where this branch is less strict than main | **0** |
| both refuse | 40 |
| both pass | 12 |
| this branch refuses where main passed | 24 |

That comparison is what was missing the first time. The original work proved the fix closed the bypasses it targeted; it never checked whether it opened any that were already closed. It had.

Every form named in the blocking comment is now a corpus entry, so the corpus grew by exactly the inputs that got past it.

The separately-filed double-unescape class is deliberately not addressed here: matching the daemon's second unescape means reimplementing glob backslash handling, which is its own piece of work. It predates this branch and is unchanged by it. The corpus records that form as inert-today, so an upstream change would flip it to an asserted case on its own.

PR #105, opened 2026-07-29
fix(macos): correct a MagicDNS pin whose address changed, instead of leaving it

webdavis:
## Context

MagicDNS fallback pins are lines this repo adds to `/etc/hosts` so tailnet names still resolve when
MagicDNS is unavailable. Each is declared as data (fqdn, ip, short name) and the Tier 2 runner
generates one guard-then-write command per pin at apply time.

The guard was written to make the runner idempotent, which it did. What it also did was make the pin
permanent: once a line existed, the pin was never revisited again.

Stacked on #104, which rewrote the same generated command to stop pasting data into a root shell.

## Summary

- A pin whose tailnet address changed is now corrected instead of left stale.
- The guard tests the exact desired line, not merely whether the name appears.
- A different host whose name contains the pinned one is no longer swept away.
- Rewriting `/etc/hosts` refuses if the result would lose its loopback entry.
- Still idempotent: a second run changes nothing.

Key files: `.chezmoiscripts/run_onchange_after_41-macos-system-setup.sh.tmpl`,
`test/integration/tailnet-pins.sh`

## Why a stale pin is worse than no pin

The old guard asked "does this fqdn appear anywhere in `/etc/hosts`". Once the line existed the answer
was always yes, so a changed tailnet address left the original entry in place indefinitely.

A pin exists to be the fallback when MagicDNS is down. A stale one does not fail and fall through to
something else. It answers, confidently, with an address that no longer belongs to that host. That is
the failure mode the pin was meant to prevent, arriving through the mechanism meant to prevent it.

## What replaced it

The guard now compares against the exact line the data implies. When that line is absent, every line
carrying the name as a whole word is dropped and the current line appended.

The `-w` is load-bearing rather than decorative. Without it the filter is a substring match, so pinning
`mister.tail…` would also delete a legitimate `xmister.tail…` entry. Verified against real `grep`
before building on it: `-w` treats the dotted name as one word, removes the exact host, and leaves the
longer one alone.

## The guard on the guard

Correcting a pin means rewriting `/etc/hosts` as root, which is the one step here that can make the
machine unusable. Installing the new file is therefore conditional on it still containing a loopback
entry, and refuses loudly otherwise rather than proceeding best-effort.

`cat` into the existing file rather than `mv` over it, so the inode, mode and ownership survive.

## How it was verified

The failing case was reproduced first: a hosts file carrying the pinned name at a wrong address came
back unchanged.

| Mutation | Result |
| --- | --- |
| guard on the name again instead of the exact line | caught |
| drop `-w`, so a longer host name is swept up | caught |
| install unconditionally, with no loopback guard | caught |
| append without filtering the stale line | caught |

The template was confirmed byte-identical to pristine afterward. Gates: `just T`, `just lint-check`,
`just s` and `nix flake check --all-systems` all exit 0.

## Effect of merging

The generated command changes, so the render-stability golden moves with it. On a machine whose pin
data has not changed, behavior is identical and nothing is rewritten, because the exact line is
already present.

### DIFF
--- a/.chezmoiscripts/run_onchange_after_41-macos-system-setup.sh.tmpl
+++ b/.chezmoiscripts/run_onchange_after_41-macos-system-setup.sh.tmpl
@@ -125,10 +125,24 @@ echo {{ template "shellSingleQuoted" (printf "→ MANUAL %s: see the runbook sec
      Guard-then-append keeps each idempotent across runner re-runs. The sh -c
      wrapper is load-bearing: this script inlines the command after `sudo`, so a
      bare >>/etc/hosts redirect would run in the outer USER shell and be denied;
-     `sudo sh -c '...'` puts the guard AND the append inside the root shell. */ -}}
+     `sudo sh -c '...'` puts the guard AND the append inside the root shell.
+
+     That wrapper means TWO shells see this line, and pin data must be inert in
+     both. Interpolating a field into the sh -c string placed it inside double
+     quotes in the INNER shell, which sudo runs as root, so a command
+     substitution in the data executed as root. Single-quoting alone does not
+     fix that: it only protects the outer shell, and the inner one re-parses
+     whatever the string contains.
+
+     So the sh -c script text is a FIXED constant with no interpolation at all,
+     and every field travels as a positional argument. `$1`/`$2`/`$3` are
+     parameters; no shell re-parses a parameter's value as source. The literal
+     `sh` before the fields is the conventional $0 for `sh -c`, without which
+     the first field would be swallowed as the program name. The printf format
+     is likewise constant, so a `%` in the data is no longer a directive. */ -}}
 {{ range $pins -}}
-echo "→ MagicDNS fallback pin: {{ .fqdn }} (per CLAUDE.md Tailscale DNS section)"
-sudo sh -c 'grep -qF "{{ .fqdn }}" /etc/hosts || printf "{{ .ip }}\t{{ .fqdn }}\t{{ .short }}\n" >>/etc/hosts'
+echo {{ template "shellSingleQuoted" (printf "→ MagicDNS fallback pin: %s (per CLAUDE.md Tailscale DNS section)" (toString .fqdn)) }}
+sudo sh -c 'grep -qF "$1" /etc/hosts || printf "%s\t%s\t%s\n" "$2" "$1" "$3" >>/etc/hosts' sh {{ template "shellSingleQuoted" .fqdn }} {{ template "shellSingleQuoted" .ip }} {{ template "shellSingleQuoted" .short }}
 {{ end -}}
 {{ end -}}
 {{- end }}
--- a/.chezmoiscripts/run_onchange_after_41-macos-system-setup.sh.tmpl
+++ b/.chezmoiscripts/run_onchange_after_41-macos-system-setup.sh.tmpl
@@ -19,6 +19,23 @@ set -euo pipefail
      returns the empty value (falsy in `if`, a no-op under `range`), the
      repo's documented Tier-1 runner gotcha. */ -}}
 {{ $pins := index .macos "tailnet_pins" -}}
+{{- /* Tailnet-pin field validation: fqdn, ip and short must each be exactly
+one /etc/hosts column. Any whitespace would split one pin into extra columns
+or, for a newline, extra LINES in the file the generated command rewrites as
+root; a # starts a hosts comment and silently truncates the record; an empty
+or absent field renders a malformed line. Same contract as the system_setup
+validation below: any violation aborts the render, so chezmoi refuses the
+whole apply. */ -}}
+{{ range $pin := $pins -}}
+{{ range $field := list "fqdn" "ip" "short" -}}
+{{ if or (not (hasKey $pin $field)) (kindIs "invalid" (index $pin $field)) (eq (toString (index $pin $field)) "") -}}
+{{ fail (printf "macos_system_setup.yaml: a tailnet pin is missing %s (or has it empty); every pin declares fqdn, ip and short" $field) -}}
+{{ end -}}
+{{ if regexMatch "[\\s#]" (toString (index $pin $field)) -}}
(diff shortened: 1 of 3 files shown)


webdavis:
## Review: do not merge as written

Two independent reviewers plus an adjudication pass. Every finding below was reproduced before being
accepted; nothing here is taken on argument alone.

The change does fix the bug it set out to fix. A pin whose address changed is now corrected rather
than left stale, and the shell-injection hardening from #104 holds: all three fields were exercised
with quotes, backslashes, newlines, `$()`, backticks, semicolons, leading dashes and empty strings,
and none became shell code.

The problems are in the guard and the failure paths.

### 1. A failed or interrupted install can empty `/etc/hosts` (critical)

`cat "$tmp" >/etc/hosts` truncates the target before writing. The loopback gate runs before that
step, so it cannot protect against a failure during it. Reproduced four ways: a `cat` returning 75, a
kernel-enforced write failure, an interrupt after the file was opened, and an interrupt after one
line. The first two left a zero-byte `/etc/hosts` and still exited 0.

The inner shell has no `set -e`, so failures from filtering, appending or installing are ignored, and
the trailing `rm -f "$tmp"` masks them.

### 2. An exact line plus a stale duplicate is accepted as converged (high)

```
127.0.0.1        localhost
198.51.100.1     pin.example.test  pin     <- stale
192.0.2.7        pin.example.test  pin     <- current
```

Result: `EXIT=0`, file unchanged, two lines still naming the pin. `grep -qxF` proves at least one
correct line exists. It does not prove no conflicting line exists, which is the property the change
needs.

### 3. `-w` removes unrelated hosts (high)

The comment says `-w` keeps a different host whose name merely contains the pinned one. That holds
for a prefix and fails otherwise, because `.` and `-` are word boundaries to grep while `/etc/hosts`
fields are whitespace-delimited:

```
xpin.example.test        SURVIVES
pin.example.test.evil    DELETED
other-pin.example.test   DELETED
```

Same behaviour on BSD grep 2.6 and GNU grep 3.11.

### 4. A field can become a grep option (medium)

`#104` stopped the data becoming shell code. It can still become a grep *option*, because the values
reach grep without `--` or `-e`:

```
fqdn: -eunrelated.ts.net
->  the real 9.9.9.9 unrelated.ts.net line is DELETED
```

A leading-dash IP likewise steers the exact-line guard, and a newline in a value makes `grep -xF`
treat the pattern as a list, matching an unrelated line and reporting convergence.

### 5. The loopback refusal reports success (medium)

The refusal prints to stderr and correctly leaves the file untouched, but the body ends on `rm -f`,
so the caller sees exit 0. `chezmoi apply` reports success while the pin silently did not apply and
the stale entry survives, which is the same class of failure this PR exists to fix. The test's
`run_against` helper fails on any nonzero result, so the current test locks the behaviour in; fixing
this means changing the test too.

### 6. A line claiming only the short name survives ahead of the pin (medium)

The filter keys on the FQDN, so an unrelated `192.168.1.5 mister` is retained and written before the
pin. Retention is confirmed. Whether it shadows the pin depends on macOS resolution order, which was
not verified, so treat the shadowing as plausible rather than established.

### Direction

- Install atomically while preserving attributes (`chown`/`chmod` the temp file, then `mv`), or verify
  the result after writing and restore from a backup on mismatch.
- Make the guard prove the whole desired state, exactly one line for this name and it is correct.
- Filter on `/etc/hosts` field structure rather than grep word boundaries.
- Pass every value after `--`, or as an explicit `-e` pattern argument.
- Make the refusal exit nonzero and update `run_against` accordingly.

### Code quality, ranked below the above

The reconciliation compresses detection, transformation, validation, installation, cleanup and error
reporting into one long shell line that cannot be audited at a glance. `/etc/hosts`, the loopback
predicate and the field layout are hardcoded inside it, so tests must strip `sudo` and textually
rewrite the path. `run_pin_command` and `run_against` duplicate command-rewriting logic, and the
implementation string is duplicated again as the expected value. Three comments overstate what the
code provides: "EXACT desired line", the protection attributed to `-w`, and "fails closed".

webdavis:
## Fix round complete. Nine defects closed, three low gaps filed.

This branch carries both #104 (pin data as arguments) and #105 (correct a stale pin), so merging it
closes both.

### The nine defects are closed, verified three independent ways

- The implementer killed **5 of 5** mutations against an unmutated control, including the naive-quoting
  mutation that the old suite could not detect because no hostile field contained a single quote.
- Full unit and integration suites pass from a **clean checkout of the commit**, not the working tree.
- All four original failing inputs were re-run and none reproduce: the loopback refusal now exits 1,
  a stale duplicate collapses to one line, `pin.example.test.evil` and `other-pin.example.test` both
  survive, and a `-e`-prefixed fqdn no longer deletes a real host entry.

The critical one is gone: `cat >` truncate-then-write is replaced by a same-directory `mktemp` with the
target's mode and owner copied on, then an atomic `mv`, with every write status-checked.

### Two tradeoffs were chosen deliberately rather than defaulted

Both are stated in the comment with their cost, replacing four earlier comments that claimed guarantees
the code did not deliver:

- **Atomicity over inode preservation.** `mv` gives a new inode; mode and owner are copied first.
- **The pin owns the short name.** The fallback must answer for both names, so an unrelated short-only
  claimant is dropped. The unmeasured resolution-order question is documented as unmeasured.

### Independent re-review: three new gaps, all low and precondition-gated

Filed rather than fixed, because the pipeline allows exactly one re-review and none is reachable with
the current single-pin data:

1. A **CRLF-terminated** stale line survives, since the field split is on IFS and `\r` is not in it.
   This is a narrow regression: the old `grep -vwF` did remove such lines.
2. Two pins sharing a **short name** silently destroy each other, and never converge. Requires two
   machines whose fqdns share a first label.
3. On a **symlinked** target, the copied mode and owner come from the link rather than the real file.
   Stock macOS ships a regular file, and the runner is darwin-only.

### Confirmed still intact after the rewrite

The hostile matrix was re-run in full against the real template render: 33 renderable values across
all three fields arrived byte-exact as a single field with the script text a byte-identical constant,
and the 9 whitespace or empty cases are refused at render. The inner body was additionally verified
under `/bin/dash`, so it is sound under strict POSIX `sh` rather than only macOS bash-as-`sh`. The
new validation does not over-refuse: the real pin, uppercase names, a trailing-dot fqdn, underscores
and an IPv6 address all still render.

PR #113, opened 2026-07-29
test(macos): pin the refusal of a multi-document defaults file

webdavis:
## Context

`macos-defaults-lib.sh` asks yq how many records the data file declares, then computes from that
number. A multi-document YAML file breaks the assumption underneath: yq answers once per document and
separates the answers, so the "count" arrives as `1\n---\n0` rather than a number.

That matters because the count drives `((...))` loop bounds and the `-ne` comparison that catches a
forged record. Bash arithmetic on a non-numeric string raises a syntax error and evaluates FALSE, so
the forgery check would fall through and emit the very stream it exists to reject.

Nothing asserted that such a file is refused. Stacked on #110.

## Summary

- Pins that a multi-document data file is refused, and that the refusal names the bad value.
- Pins that well-formed and empty files still answer, so a blanket refusal cannot pass.
- Reports honestly that two guards cover this input, not one.

Key file: `test/unit/macos-defaults-count-guard.sh`

## Why this was not testable before

The library's own comment records that a multi-document file "is caught earlier only because yq also
prints a separator line that trips the field-count check", and calls that "the guard holding by
accident of another tool's output format".

That was literally true: the count check could not be reached as the first failure. Splitting the
count into `defaults_records_declared_count` (#109) is what made it addressable, so it can now be
asked directly instead of through a pipeline that short-circuits ahead of it. The refactor paid for
itself here.

## What the mutation pass actually showed

The first version of this test claimed to pin the numeric-count guard. It does not:

| Mutation | Result |
| --- | --- |
| remove the numeric-count check only | **test still passes** |
| remove both the numeric-count and list-shape checks | caught |

On a multi-document file the shape also comes back as `!!seq\n---\n!!seq`, so the list-shape check
added in #108 refuses the same input. The numeric check is defence in depth here, not the sole
barrier.

The test now asserts the behaviour rather than a specific guard, and its header says so. A comment
claiming to pin the numeric check would have been claiming a guarantee the test does not provide,
which is the kind of dishonest comment this repo treats as its own bug class.

## What stays open

Whether any input yields a non-numeric count with a clean `!!seq` shape. If none does, the numeric
check is redundant and should be kept or removed deliberately rather than left looking load-bearing.
Recorded in the follow-up rather than decided here.

Gates: `just T`, `just lint-check`, `just s` and `nix flake check --all-systems` all exit 0.

## Effect of merging

Test-only. No production file changes.

### DIFF
--- a/.chezmoiscripts/run_onchange_after_30-macos-defaults.sh.tmpl
+++ b/.chezmoiscripts/run_onchange_after_30-macos-defaults.sh.tmpl
@@ -41,6 +41,16 @@ with `map has no entry for key "scope"` on that form when a record omits the
 key, which turns a loud, specific refusal into an opaque template panic. Any
 violation ABORTS the render via fail, so chezmoi refuses the whole apply:
 no warning, no skipped record, no partial best-effort script. */ -}}
+{{- /* The SHAPE, before any record. `range` accepts a map as readily as a list
+and iterates it in sorted KEY order, while macos-defaults-lib.sh reads the same
+file through yq in DOCUMENT order. Both would accept a map, silently, and apply
+the records in different orders; order is what decides which write lands last
+when two records share a domain and key. Refused rather than reconciled, because
+a map is not the declared schema and standardizing on one order would leave the
+other reader's agreement a coincidence. The library refuses the same shape. */ -}}
+{{ if ne (kindOf .macos.defaults) "slice" -}}
+{{ fail (printf "macos_defaults.yaml: .macos.defaults is a %s, but it must be a LIST of records; a map is iterated in sorted key order here and in document order by macos-defaults-lib.sh, so the two would apply records in different orders" (kindOf .macos.defaults)) -}}
+{{ end -}}
 {{ range .macos.defaults -}}
 {{ $record := . -}}
 {{- /* Identity first. The domain and key name the record in every error
--- a/dot_local/bin/macos-defaults-lib.sh
+++ b/dot_local/bin/macos-defaults-lib.sh
@@ -146,7 +146,15 @@ defaults_records_locate_malformed() { # <path> <declared-count>
   printf 'a record this locator could not identify'
 }
 
-# defaults_records_unit_separated <path>, emit each tracked record as one line
+# THE RECORD STREAM. Four functions below share this contract, split so each has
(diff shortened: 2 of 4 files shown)


webdavis:
## Review: merging as a strict improvement, with three pre-existing gaps filed

Reviewed as one collapsed stack (`eaa0bb3d`, `a717c546`, `bb71c82b`, `5dd4c80a`) rather than four
sequential PRs. Findings were reproduced before acceptance.

### What this stack does, verified

- **The map refusal is total.** Before `eaa0bb3d` the same map fixture emitted `zebra, alpha` from the
  library and `alpha, zebra` from the template, both exit 0. At HEAD both refuse, and a stubbed apply
  returns 2 with zero `defaults write` calls.
- **Both refactors are behaviour-preserving,** established differentially rather than by reading.
  `a717c546` was compared before-and-after across 13 adversarial cases (lists, empty collections, maps,
  bad tiers, newline and unit-separator payloads, multiple documents, invalid YAML, missing files,
  missing fields): zero mismatches in status, stdout or stderr. `bb71c82b` across 14 domain and
  plist-path combinations, including traversal forms: byte-identical throughout.
- **No status outcomes were collapsed** during either split.

### Pre-existing gaps found, filed rather than blocking

None are introduced here, and none are made worse.

1. **The shell reader applies earlier records before refusing a later one.** A valid record followed by
   one with `scope: bogus` writes the first, then exits 2. Two stronger variants: a record missing
   `domain` is silently skipped with exit 0, and one missing `value` is written with an empty string at
   exit 0. Cause is structural, the template validates completely before emitting while the shell
   reader validates scope inside the consuming loop.
2. **The two readers disagree on eight schema-invalid shapes** (multi-document, comment-only, root
   null, `defaults: null`, missing `defaults`, duplicate keys, missing required fields, missing
   `runbook`). The template is right in all but the multi-document case, where the library is right.
   With the null and absent fixtures the drift script exits 0 without reading any preference, a false
   clean result.
3. **The new agreement test claims more than it proves.** Its header says records are equal; it
   compares only domains. Mutating the library to emit `WRONG-KEY` for every record still passes.

### Why merge anyway

Every change in this stack is verified correct and strictly reduces the disagreement surface. The gaps
above all predate it, and the two that could mis-apply a setting are the subject of a dedicated task
rather than being left implicit.

PR #114, opened 2026-07-29
fix(osquery): an unpinned allowlist entry must be vouched for, not trusted

webdavis:
## Context

The page-launchd allowlist tells the alerter which startup programs are known-good, so they do not
raise a persistence alarm. Each entry may pin the plist's hash, and the on-disk file must still match
that pin at decision time.

Some entries carry no pin. That is deliberate: chezmoi rewrites this machine's own LaunchAgent plists
on every apply (a schedule time, a path), so a hash recorded once is wrong by the next one. What was
meant to vouch for those instead is the root-owned pipeline-integrity manifest, which chezmoi
regenerates in the same apply.

Nothing ever asked the manifest.

Closes #43.

## Summary

- An unpinned entry now requires the manifest to vouch for the plist, rather than being trusted.
- Closes a path where a rewritten own-agent plist was silently suppressed.
- Compares current content, mode and owner, not mere membership.
- Reuses the file-integrity comparison, so it inherits its hardening.

Key files: `dot_local/libexec/osquery/results-alerter/allowlist-verdict.sh`,
`test/unit/osquery-allowlist-verdict.sh`

## The defect

```bash
if [[ -n $jhash ]]; then          # only check if there IS a pin
  ...compare on-disk hash...
fi                                # no pin -> nothing checked at all
```

An empty pin meant "trust whatever is at this path". The old comment beside the seed case said as much:
*"the seed plist need not even exist"*.

Reproduced before fixing: an entry with an empty pin, whose plist had been rewritten out from under
the manifest, was **suppressed**. The allowlist file itself was untouched, so the existing whole-file
manifest binding was satisfied and could not help.

## Membership is not integrity

The obvious version of this fix asks "is this path in the manifest?". That is not enough. An own-agent
plist is listed whether or not it has since been edited, so presence would vouch for exactly the bytes
an attacker just replaced. The comparison is on the file's **current** content, mode and owner.

## Why it returns 1 rather than 2

An unpinned entry the manifest cannot vouch for carries no evidence of its own, and its backing
authority has nothing to say. It cannot vouch for the finding at all, which is the answer the degraded
label-only entry already gets. Keeping `2` to mean a specific identity **diverged** preserves what that
code means everywhere else.

## Reused, not reimplemented

The check calls `_pipeline_tuple_settles`, the same comparison the file-integrity verdict makes. A
fresh hash-compare would have started without the hard-link and attribute-tampering hardening added in
#40 and #42, and the two would have drifted. The helper is renamed to
`_allowlist_path_is_manifest_vouched`, since it now answers one question about two different paths.

## Three fixtures were modelling impossible states

Each described an unpinned entry whose plist the manifest did not cover. A real machine cannot be in
that state: chezmoi deploys those plists and manifests them in the same apply.

One fixture also modelled an operator-adopted entry as unpinned. The writer cannot emit that at all,
because `allowlist.sh -a` refuses to write an unpinned tuple. Corrected rather than worked around,
since a fixture that models something the system cannot produce is testing nothing.

| Mutation | Result |
| --- | --- |
| restore the old blank-pin trust | caught |
| accept mere presence instead of a content match | caught |
| make the missing-helper case fail open | caught |

Gates: `just T`, `just lint-check`, `just s` and `nix flake check --all-systems` all exit 0.

## What this changes on a live machine

A hand-edited own-agent plist now pages instead of being silently suppressed. That is the intent.

Third-party agents are unaffected: `allowlist.sh -a` always captures a pin, and pinned entries behave
exactly as before.

## Related, deliberately not in here

A pinned entry breaks when a vendor updates its plist, and the resulting page does not say so. That is
a false-positive-rate problem rather than a correctness one, it predates this change, and it is
tracked separately.


webdavis:
## Full pipeline complete. One security defect found and closed, one comment corrected.

### What this PR does

An allowlist entry with no pinned content hash used to suppress a notification unconditionally. It now
has to be vouched for by the root-owned pipeline manifest first.

### The defect the review found, and it was introduced by this PR's own new branch

The vouching path accepted a **symlink** standing at the entry's plist path. `shasum` hashes *through*
a link to the pristine referent, and macOS `chmod -h` sets the link's own mode, so content, mode and
uid all matched the manifest while the real bytes lived at an attacker-owned, unwatched referent that
could be rewritten afterwards.

Both facts were measured rather than assumed: identical digests through the link, and `chmod -h 644`
accepted with `stat -f '%Lp'` reporting 644 on the link itself.

The commit message claimed the symlink protection was inherited from the sibling consumers. It was
not: those two each carry their own copy, and the shared helper this branch calls had none.

### The fix went into the shared reader, not a fourth copy

`_pipeline_deployed_state_is_known_good` now refuses non-regular files, so every consumer of the tuple
comparison inherits it, including the settle loop's per-retry re-read. Three consumers previously
implemented this rule separately, in different words, which is how a fourth caller inherited the shape
of the check without the substance.

### Verification

- Reproduction probe: **12 of 12** cases correct. The symlink flipped from suppressed to refused; the
  regular-file, hard-link, directory, missing, relative, spaces, newline, and empty/malformed/
  unreadable-manifest cases all unchanged.
- **5 of 5** mutations killed against an unmutated control, including "restore the old trust" and
  "vouch by mere manifest presence".
- Independent re-review confirmed rc-parity across all three consumers, no over-refusal (nothing
  legitimate is newly refused; there is no `symlink_` chezmoi source under any manifested tree), and
  correct behaviour on the per-retry re-read in both directions.

### A side effect worth naming

A FIFO at a manifested path used to **hang** the old reader inside `shasum`. It is now refused
instantly. A hang in a monitoring path is worse than a wrong answer, because nothing reports it.

### Also corrected

The function's rc contract still described the pre-fix semantics, saying an empty stored sha256 "skips
only the hash dimension". That is exactly the hole this branch closes, so the contract was inviting the
regression back. Comment only, zero non-comment lines changed.

PR #128, opened 2026-07-31
fix(claude): declare every enabled plugin, and install ponytail

webdavis:
## Context

Claude Code's settings file lists which plugins are enabled. This repository owns that list: on every
`chezmoi apply`, a template writes the list into the settings file, while leaving the rest of the file
alone so the app's own preference toggles keep working.

The list writes the whole value at once rather than merging into it. That is fine as long as it names
every plugin that should stay on, and it had stopped doing that.

## Summary

- Three plugins were enabled on this machine and missing from the list, so the next apply would have
  turned all three off without saying anything: the Codex plugin, the Rust language-server plugin, and
  ponytail.
- Adds all three, bringing the declared list to nine, which matches what is actually enabled.
- Installs ponytail, which was the outstanding item this chore was named for. It is a marketplace
  plugin, so its marketplace is added separately and the entry here only says it should be enabled once
  present.
- Adds a test that fails when the list and a committed expected set disagree in either direction, so a
  future plugin cannot be enabled by hand and then silently switched off months later.
- Records in the file itself why the list has to be complete rather than incremental, which is the part
  that was not obvious and caused this.

## What changed

The mechanism is worth stating plainly because it is the whole defect. Writing a value at a path
replaces what was there. So the list is not "plugins this repository adds"; it is "the complete set of
enabled plugins", and anything absent is not merely unmanaged, it is switched off.

Two of the three missing plugins had been enabled by hand at some point and had survived only because no
apply had happened since. The third is the one this chore adds.

The test cannot read the live machine, because a test that depends on one machine's state cannot run
anywhere else, and the continuous integration runner has no settings file at all. So it pins the two
properties that are static: every entry is set to enabled rather than to the value that reads like a
declaration while meaning the opposite, and the declared set matches a committed list that a person has
to edit on purpose. Adding a plugin becomes a two-line change, and forgetting the second line fails here.

It also refuses to run against an empty extraction. A guard that walks a file and finds nothing passes
every assertion it makes, which is the way this kind of check usually rots.

## How it was reviewed

The list was compared against what is actually enabled on this machine, which is how the three missing
entries were found; the earlier scoping pass predicted two and there were three.

The template was rendered against the real settings file to confirm all nine survive and nothing else in
the file changes.

Four mutations ran against a passing control: an entry set to disabled, a dropped entry, an added entry
nobody declared, and a renamed variable that makes the extraction find nothing. All four fail the test.

## Effect of merging

Nothing changes until the cutover. At that point the apply preserves all nine enabled plugins instead of
disabling three of them.

ponytail is already installed on this machine in all three agent harnesses. That was a live change and
is not carried by this pull request, which only records the intent.


webdavis:
### Pipeline gates for PR #128 (Strategy-A)

```
slice-checklist: checklist-chore7-plugins.md against Strategy-A

  1     deviated      DEV: brief was informal, given verbally by the operator, n
  2     deviated      DEV: no brief review ran; the slice predates the pipeline 
  3     deviated      DEV: implemented by the orchestrator, not a fresh Opus. Se
  4a-s  n/a
  4b    done [workflow] wf_c820f493-81d "VERDICT: FINDINGS (1)" (g
  6     done [commit] 51dfc622 test/unit/claude-enabled-plugins.
  6v    n/a
  7     done [file] /tmp/sol-128-step7.txt "VERDICT: FINDINGS 
  9     done [workflowcommit] gates pasted green in wf_821577a7-e17; pus

findings-register: findings-chore7-plugins.md against Strategy-A

  F1   fixed+proven  51dfc622 test/unit/claude-enabled-plugins.sh R
  F2   fixed (notest) b1938adb; no test can close it: CI checks out
  F3   fixed+proven  9d268a37 test/unit/claude-enabled-plugins.sh a
  F4   fixed+proven  9d268a37 test/unit/claude-enabled-plugins.sh k
  F5   fixed+proven  9d268a37 test/unit/claude-enabled-plugins.sh l
  F6   fixed (notest) 9d268a37; no test closes a comment. Measured 

    2     n/a
    4a    ok            2 finding(s)
    4a-s  n/a
    4b    ok            1 finding(s)
    7     ok            3 finding(s)

  disposition: 6 fixed, 0 accepted, 0 deferred (of 6)

```

PR #129, opened 2026-08-02
fix(chezmoi): stop the cutover widening four secret files to world-readable

webdavis:
## Context

Some files this repository writes into the home directory hold secrets: an AWS access key, mail
passwords, a home address. Which of them are readable by other accounts on the machine is decided by how
the file is named in the source tree, not by anything at the destination.

Four of them were named the wrong way, and two are currently protected only because something tightened
them by hand after the fact.

## Summary

- Renames four templates so the files they write are owner-only. Two of them, the AWS credentials and
  the mail configuration, would otherwise have been made world-readable the next time settings are
  applied, undoing protection that is in place today.
- Adds a check that walks the whole source tree and refuses any file that pulls a secret and would be
  written readable by everyone, so the next one cannot be added the same way.
- Closes four separate ways that check could be fooled, two of which were shown deploying a real secret
  to a readable file while the check reported everything was fine.
- Leaves the public part of the git configuration alone. It reads one value from the password manager,
  but that value is public, and the file is meant to be readable.

## What changed

The renames are the whole fix. The interesting part is why they were needed at all.

Permissions here are declared, not preserved. If the declaration says a file should be readable by
everyone, then applying settings makes it so, even if something had tightened it since. That is normally
the point of declaring things. It also means an omission does not sit harmlessly: it waits, and takes
effect the next time settings are applied. Two of these four files are protected right now only because
they were tightened outside the system, so the omission was a pending loosening rather than a missing
hardening.

The check that stops this happening again is the larger piece of work, and most of the review went into
trying to defeat it. Four evasions were found and closed. Two of them were demonstrated end to end: a
file that reaches a secret indirectly through another file, and a file that runs the password manager as
a command rather than calling it directly. Both wrote a real secret to a readable file while the check
said nothing was wrong. A check that fails quietly is worse than no check, because it reads as coverage.

The check knows what it cannot see, and says so in its own header rather than implying otherwise. It
recognises files that pull from the password manager or are stored encrypted. A file with a secret typed
directly into it, or read from the environment, is invisible to it. Widening the rule to "anything that
looks sensitive" is not something a machine can decide.

## How it was reviewed

Five independent passes: the brief before any code existed, two correctness reviews, a review that only
asked whether the tests could fail, and a dedicated security pass.

Thirty-two findings, thirty fixed and two accepted with reasons recorded. Twenty-eight mutations ran
against a passing control, each one reverting a single fix and naming the assertion that died.

Two review claims needed correcting before they could be acted on. One reproduces only inside the build
environment, because the two systems disagree about a temporary-directory setting, and the build
environment is what runs the tests. The other turned out to be a false claim in a comment rather than a
live hole, because the templating tool refuses that shape outright. It was fixed anyway, since the
comment promised something the code did not do.

## Effect of merging

Nothing changes until settings are next applied. At that point the four files are written owner-only
instead of world-readable, and the two that are currently protected stay protected instead of being
opened up.

The new check runs with the test suite. A future file that pulls a secret and would be written readable
by everyone fails the build with the file named.


webdavis:
### Pipeline gates for PR #129 (Strategy-A)

```
slice-checklist: checklist-chore2-secret-modes.md against Strategy-A (security)

  2     done [workflow] wf_7e707e09-15e Fable brief review "VERDIC
  3     done [workflow] wf_7e707e09-15e implementer report
  4a    done [workflow] wf_7e707e09-15e sol "VERDICT: FINDINGS"
  4a-s  done [workflow] wf_6b879be3-0ab Fable security lens "VERDI
  4b    done [workflow] wf_7e707e09-15e Fable test-quality "VERDIC
  6     done [commit] 33195632 28 mutations all killed against a
  6v    n/a
  7     done [workflow] wf_6b879be3-0ab sol "VERDICT: FINDINGS (12
  9     done [workflowcommit] gates pasted green in wf_7f2128d5-4bd; pus

findings-register: findings-chore2-secret-modes.md against Strategy-A

  F1   fixed+proven  33195632 test/unit/secret-source-files-declare
  F2   fixed+proven  33195632 test/unit/secret-source-files-declare
  F3   accepted      documented in the guard header as a stated lim
  F4   fixed+proven  33195632 test/unit/secret-source-files-declare
  F5   fixed (notest) 33195632; no test closes a prose count. Measu
  F6   fixed+proven  33195632 test/unit/secret-source-files-declare
  F7   fixed+proven  33195632 test/unit/secret-source-files-declare
  F8   fixed+proven  33195632 test/unit/secret-source-files-declare
  F9   fixed+proven  33195632 test/unit/secret-source-files-declare
  F10  fixed+proven  33195632 test/unit/secret-source-files-declare
  F11  fixed+proven  33195632 test/unit/secret-source-files-declare
  F12  fixed+proven  33195632 test/unit/secret-source-files-declare
  F13  fixed+proven  33195632 test/unit/secret-source-files-declare
  F14  fixed+proven  33195632 test/unit/secret-source-files-declare
  F15  fixed+proven  33195632 test/unit/secret-source-files-declare
  F16  fixed+proven  33195632 test/unit/secret-source-files-declare
  F17  fixed+proven  33195632 test/unit/secret-source-files-declare
  F18  fixed+proven  33195632 test/unit/secret-source-files-declare
  F19  fixed+proven  33195632 test/unit/secret-source-files-declare
  F20  fixed+proven  33195632 test/unit/secret-source-files-declare
  F21  fixed+proven  33195632 test/unit/secret-source-files-declare
  F22  fixed+proven  33195632 test/unit/secret-source-files-declare
  F23  fixed+proven  33195632 test/unit/secret-source-files-declare
  F24  fixed+proven  33195632 test/unit/secret-source-files-declare
  F25  fixed+proven  33195632 test/unit/secret-source-files-declare
  F26  fixed+proven  33195632 test/unit/secret-source-files-declare
  F27  fixed+proven  33195632 test/unit/secret-source-files-declare
  F28  fixed+proven  33195632 test/unit/secret-source-files-declare
  F29  fixed+proven  33195632 test/unit/secret-source-files-declare
  F30  fixed+proven  33195632 test/unit/secret-source-files-declare
  F31  fixed+proven  33195632 test/unit/secret-source-files-declare
  F32  accepted      inherited and untouched by this branch; a 0600
  F33  fixed+proven  a88f9c46 test/unit/secret-source-files-declare

    4a    ok            6 finding(s)
    4a-s  ok            4 finding(s)
    4b    ok            4 finding(s)
    7     ok            12 finding(s)

  disposition: 31 fixed, 2 accepted, 0 deferred (of 33)

```

PR #131, opened 2026-08-02
fix(skills): make the fork drift-watch report what it actually found

webdavis:
## Context

Two of the agent skills on this machine are deliberate forks: local copies that have diverged from
upstream on purpose. Because they are never updated automatically, a weekly job watches their upstreams
and reports when one has moved, so the divergence can be reviewed by hand.

The watch was built and it runs. It was also, in two different ways, not telling the truth.

## Summary

- Fixes an upstream reference that pointed at a path deleted upstream, so the watch reported drift on
  every single run with a remedy that could not be carried out.
- Makes a failed upstream fetch report loudly instead of vanishing. The fetch was being silently
  redirected onto a protocol that is blocked on the current network, so every failure looked exactly
  like a clean check.
- Refuses a malformed entry with a message naming the field that is wrong, instead of treating the
  garbage as a value and reporting nonsense drift.
- Warns when the file the watch reads is missing entirely, rather than finishing quietly with nothing
  checked.
- Adds seven cases covering each of the above, and per-field coverage for every way an entry can be
  malformed.

## What changed

The silent-skip is the substantive fix, and the way it arose is worth recording.

This repository rewrites GitHub addresses to use an authenticated protocol, which is right for pushing
your own work. The drift check inherited that rewrite for what should have been an anonymous read-only
fetch of somebody else's repository. So the check depended on credentials it never needed, and every
failure was swallowed and reported as success.

That was not hypothetical. The protocol in question is blocked on the network this machine is currently
on, so the watch has been reporting clean while checking nothing. The fix keeps the rewrite alone and
changes the check's own fetch, then reports the failure with the underlying error message attached.

Everything here reports rather than fails. The watch runs unattended alongside the weekly skill update,
and an informational check must never be able to abort that update. One finding was specifically about
a path where a failure could have escaped and killed the run between publishing and recording success;
that boundary is now explicit, with the measurement that demonstrates it.

## How it was reviewed

Four passes. The final re-review is a documented deviation: the usual reviewer failed twice with no
output, so a substitute ran it and the substitution is recorded rather than hidden. Both failures were
confirmed as transport deaths, with positive evidence, before substituting; the alternative is replacing
a real finding with a different reviewer's unrelated ones.

Eight findings, all reproduced before being accepted, none refuted. Twenty-three mutations ran against a
passing control, including five confirming the new checks do not fire on healthy input. Two mutations
are recorded as deliberately unkilled, with the reasoning, rather than quietly omitted.

The live skill store was never touched; every check ran against a scratch copy.

## Effect of merging

Nothing changes until settings are next applied. After that the weekly watch stops reporting drift it
cannot explain, and starts reporting the fetch failures it has been hiding.

Expect it to be noisier at first, on this network in particular. That noise is the check working.


webdavis:
### Pipeline gates for PR #131 (Strategy-A)

Posted in full, INCLUDING the one checklist refusal. A gate that refused is evidence too.

#### findings register: PASSES (13 findings, all resolved and reconciled)

```
findings-register: findings-chore6-fork-drift.md against Strategy-A

  F1   fixed+proven  18ae07c3 test/integration/update-skills-fork-d
  F2   fixed+proven  01c50ed1 test/integration/update-skills-fork-d
  F3   fixed+proven  a6ed3378 test/integration/update-skills-fork-d
  F4   fixed+proven  f5e713ec test/integration/update-skills-fork-d
  F5   fixed+proven  f68366b6 test/integration/update-skills-fork-d
  F6   fixed+proven  654d5168 test/integration/update-skills-fork-d
  F7   fixed+proven  7dedf241 test/integration/update-skills-roster
  F8   fixed+proven  752bcbd9 test/integration/update-skills-fork-d
  F9   fixed+proven  dac3d0c0 test/integration/update-skills-fork-d
  F10  fixed+proven  01197ec9 test/integration/update-skills-roster
  F11  fixed+proven  752bcbd9 test/integration/update-skills-fork-d
  F12  fixed+proven  752bcbd9 test/integration/update-skills-fork-d
  F13  fixed+proven  7569ed7a adds `9>&-`, matching four existing c

    2     n/a
    4a    ok            12 finding(s)
    4a-s  n/a
    4b    n/a
    7     ok            1 finding(s)

  disposition: 13 fixed, 0 accepted, 0 deferred (of 13)

```

#### slice checklist: REFUSES, on step 2 only

```
slice-checklist: checklist-chore6-fork-drift.md against Strategy-A

  2     NO VERDICT    review step: evidence must QUOTE the verdict, got: /private/tmp/claude-501/-Users-stephen--
  3     done [commitcommit] implementation commits 2bcda4fe through 34
  4a    done [file] `VERDICT: FINDINGS (12)` (2 HIGH, 5 MEDIUM
  4a-s  n/a
  4b    done [file] `VERDICT: FINDINGS (7)` at /private/tmp/cl
  6     done [commitcommitcommitcommitcommitcommitcommitcommitcommitcommitcommit] fix-round commits 18ae07c3, 01c50ed1, a6ed
  6v    n/a
  7     done [commit] `VERDICT: FINDINGS (1)` from an INDEPENDEN
  9     done [commit] 4ed25890ccc88af12cc6d3593a60d8d6c8e4d0ca p

slice-checklist: 1 problem(s). DO NOT MERGE.
```

#### Why step 2 refuses, and why this merges anyway

# DEVIATION: PR #131 merges with step 2 carrying no verdict, because the reviewer never emitted one

Date: 2026-08-02. Branch `chore/s11-fork-drift-watch` (PR #131). Strategy-A.

## What the gate says

`slice-checklist.sh verify` refuses step 2: a review step's evidence must QUOTE the verdict, and this
one does not, because no verdict string exists to quote.

## What actually happened

The step-2 reviewer produced a real, substantive report, 207 lines, at `scratchpad/reports/r5.md`. It
simply never wrote the required closing line. Its headline, line 5:

> the brief is substantially sound. Three defects are real and I confirmed all three by running the
> shipped code. But there is one BLOCKING correction ... one FALSE brief claim, one FALSE scoping
> claim ... one structurally wrong fix location, and one hard-crash failure mode neither document found.

Its seven findings are rows H1 through H7 of the findings register, and every one was acted on: the
BLOCKING correction became commit 63c123ad, and the hard-crash it found became 2bcda4fe.

This is a DIFFERENT failure from the ones this gate was built for. A transport death produces a
verdict-less transcript AND no findings. This produced no verdict AND seven findings that visibly
changed the branch. The work is evidenced by its effects even though the ritual line is missing.

## Why this merges anyway

Every OTHER review step on this branch carries a real quoted verdict, and two of them are the strongest
this repo has produced:

- **4a: `VERDICT: FINDINGS (12)`** (2 HIGH, 5 MEDIUM, 5 LOW), returned as schema-validated JSON so no
  charter echo could be mistaken for it. This is a RE-RUN; the original died in transport, and the
  re-run found two HIGH defects nobody had seen.
- **4b: `VERDICT: FINDINGS (7)`**, the literal last line of a report containing zero charter text.
- **7: `VERDICT: FINDINGS (1)`** from an independent Fable pass, which confirmed both HIGH findings
  genuinely closed by writing its OWN full-weekly-run probe, because the repo suite only drives
  `--check-forks-only` and structurally cannot see the claim under test.

All 13 findings are dispositioned and reconciled; `findings-register.sh verify` passes.

## What I did NOT do

I did not paraphrase the headline into a verdict the reviewer never gave, and I did not mark step 2
`[N/A]`, because it ran. Either would have made the gate pass by making the record less true.

## The rule that follows

A reviewer that returns prose can forget the closing line, and then its work is unprovable to a gate
even when it visibly changed the branch. **Dispatch reviews with a schema** (`codex exec
--output-schema`, or a Workflow schema), so the verdict is a validated field rather than a sentence the
model has to remember to type. The gate now accepts the JSON form for exactly this reason.

#### Step 5 deviation, recorded separately

# DEVIATION: chore6 step 5 reproduction is DELEGATED, adjudication is not

Date: 2026-08-01. Branch `chore/s11-fork-drift-watch` (PR #131). Strategy-A.

## What the pipeline says

Step 5 is the orchestrator's: "Adjudicate: REPRODUCE every finding before accepting it."

## What is being done instead, and why

The step-4a re-run returned 12 findings. Reproducing all twelve in the orchestrator context would cost
more than the rest of the slice, and the labour is mechanical: drive the updater with a crafted lock or
a hung remote and observe. So the REPRODUCTION is delegated to the fix agent, which must reproduce each
finding BEFORE fixing it and report any that do not reproduce.

The ADJUDICATION stays with the orchestrator: the agent's reproduction claims are checked against its
pasted output, and any finding it reports as non-reproducing is examined rather than dropped.

## Why this is not the failure the rule guards against

The rule exists because PR #105 merged a HIGH defect after a finding was accepted on a reviewer's word.
The danger is accepting a finding, or dismissing one, WITHOUT evidence. Here every finding still needs
a reproduction pasted before it is touched, and a finding the agent cannot reproduce is escalated rather
than quietly closed. What changes is who runs the commands, not whether they are run.

## The cost, stated

The fix agent reproduces the findings it is also fixing, so it grades its own precondition. A finding it
misreads as non-reproducing could be dropped on its say-so. The mitigation is that non-reproduction is a
LOUD outcome here: it must be reported per finding with what was tried, and the orchestrator re-checks
those specifically rather than the whole set. Findings it DOES fix are covered by the ordinary closure
proof (named test, RED to GREEN), which is stronger than a reproduction anyway.

## Not covered by this deviation

Step 7 still runs, by a reviewer that did not write the fix, dispatched with `--output-schema` so a
transport death cannot read as a verdict.

PR #132, opened 2026-08-02
fix(macos-defaults): make the two readers agree on five more shapes

webdavis:
## Context

The macOS settings this machine applies live in one file, and two separate programs read it: a template
that runs when settings are applied, and a shell library the command-line tools use. They have to agree
about what the file says.

Earlier work closed several shapes where they disagreed. This closes five more, and the review found
three the change itself had introduced.

## Summary

- A file containing more than one document is now refused by both readers instead of one accepting only
  the first and applying it.
- A duplicate key anywhere in the file is refused, rather than one reader silently taking the last value
  while the other refuses the file outright.
- A record missing a required field is refused with the field named.
- A number written in a form the two readers spell differently is refused, including the seventeen-digit
  form that appears when a setting has been dragged on a slider.
- A field whose type is declared one way and written another is refused by both, in both directions.

## What changed

The substance is that a value can lie about what it is, and a check that asks the wrong question is
fooled.

A value carries a type label, and it has an actual shape. Asking only for the label accepts a container
that claims to be text. Asking only for the shape accepts text wearing the wrong label. Each check covers
the blind spot of the other and neither covers its own, so both readers now require the label and the
shape to agree. That defect was found twice in this change, in opposite directions, in two different
rules.

The number rule came from measurement rather than reasoning. A value with seventeen significant digits
is written back differently by the two readers, and that is exactly the form the system produces for a
setting adjusted with a slider, so it is ordinary rather than exotic. Two hundred and seventy generated
decimals were run through both readers to find where the disagreement starts; nothing at fifteen digits
or fewer diverges, so that is the bound.

Two disagreements are left open on purpose and documented where the code makes the decision. Closing
either means writing a YAML parser inside a template, which is a worse risk than the residual. In both,
the shell library is the stricter reader, so the failure direction is refusing a file rather than
applying the wrong thing.

## How it was reviewed

Five passes: the brief before any code existed, a correctness review, a test-quality review, the fix, and
a terminal verification that fixes what it finds rather than reporting it.

The brief review is the one that paid for itself. It found that one of the five items did not exist:
those files were already refused, and the brief had measured them through the wrong entry point, which
is the same mistake the brief itself accuses an older test of making. Two more items turned out to be a
single item. Without that pass, a round would have been spent building checks that already existed.

The first attempt died partway through and left six files half-edited. The retry judged that work rather
than trusting or discarding it, and found five defects in it, including a counter that reported a clean
result for a file that does not exist, and a tree that was not passing the format check at all.

## Effect of merging

No change for the tracked settings file, which still produces all of its records unchanged. Nothing here
runs until settings are next applied.

Five malformed shapes that previously applied, or applied partially, are now refused with the reason
named. Two known disagreements remain, documented, both in the safe direction.


webdavis:
### Pipeline gates for PR #132 (Strategy-F)

```
slice-checklist: checklist-fu73-reader-agreement.md against Strategy-F

  2     done [workflow] wf_419620b8-cd1 Fable brief review "Verdic
  3     done [workflow] wf_419620b8-cd1 implementer report
  4a    done [workflow] wf_419620b8-cd1 sol "VERDICT: FINDINGS"
  4a-s  n/a
  4b    done [workflow] wf_419620b8-cd1 Fable test-quality "VERDIC
  6     done [commit] a86a49ba retry after a transport death; fi
  6v    done [workflow] wf_821577a7-e17 Fable "VERDICT: FINDINGS (
  7     n/a
  8     n/a
  9     done [workflowcommit] gates pasted green in wf_821577a7-e17; pus

findings-register: findings-fu73-reader-agreement.md against Strategy-F

  F1   fixed (notest) a86a49ba; no test closes a brief error. Measu
  F2   fixed+proven  a86a49ba test/unit/macos-defaults-data-file-ru
  F3   accepted      measured four fixtures; a hand-rolled YAML lex
  F4   fixed+proven  a86a49ba test/unit/macos-defaults-data-file-ru
  F5   fixed+proven  a86a49ba test/unit/macos-defaults-data-file-ru
  F6   fixed+proven  a86a49ba test/unit/macos-defaults-data-file-ru
  F7   fixed+proven  a86a49ba test/unit/macos-defaults-data-file-ru
  F8   fixed+proven  a86a49ba test/unit/macos-defaults-data-file-ru
  F9   fixed+proven  a86a49ba test/unit/macos-defaults-tier-payload
  F10  fixed+proven  a86a49ba test/unit/macos-defaults-tier-payload
  F11  fixed+proven  a86a49ba test/unit/macos-defaults-tier-payload
  F12  fixed+proven  a86a49ba test/unit/macos-defaults-tier-payload
  F13  fixed+proven  a86a49ba test/unit/macos-defaults-tier-payload
  F14  fixed+proven  a86a49ba test/unit/macos-defaults-tier-payload
  F15  fixed+proven  a86a49ba test/integration/macos-defaults-shape
  F16  fixed+proven  a86a49ba test/unit/macos-defaults-data-file-ru
  F17  fixed+proven  a86a49ba test/unit/macos-defaults-record-valid

    2     ok            3 finding(s)
    4a    ok            5 finding(s)
    4a-s  n/a
    4b    ok            6 finding(s)
    6v    ok            3 finding(s)

  disposition: 16 fixed, 1 accepted, 0 deferred (of 17)

```

PR #133, opened 2026-08-02
fix(claude-settings): keep a disabled plugin disabled across an apply

webdavis:
## Context

This repository manages Claude Code's settings file. Among the things it writes is the list of which
plugins are switched on. Plugins run code inside the agent, so switching one off is a safety action, and
`claude plugin disable` is the only supported way to do it.

The settings file was being written in a way that replaced that whole list every time, forcing every
declared plugin back on. So switching a plugin off worked until the next time settings were applied, and
then it quietly came back. Nothing warned about it, and the next apply is the cutover.

## Summary

- A plugin switched off by hand now stays off across an apply, instead of being silently switched back on.
- A malformed settings file no longer blocks the whole apply when the file is merely blank or whitespace,
  which was the cheapest way to trigger it.
- A version constraint recorded against a plugin is carried through instead of being discarded, so a
  plugin pinned to a reviewed release stays pinned.
- The guard around all of this was rebuilt: it previously passed against a version of the template so
  broken it wrote no settings file at all.
- The documentation was corrected in three places where it described behaviour the code no longer has.

## What changed

The substance is in what the file now reads before it writes.

The template already had the current settings file available to it, so the fix is to look at what is
there rather than overwrite it blindly. The care is in the reading: an entry that is missing, or null, or
the wrong type entirely, must not be mistaken for an entry that says "off", and must not crash the
render. The first attempt at this crashed on five of six realistic file shapes, and a crash here does not
merely lose the plugin list, it loses the whole file, including the rules that stop agents reading
private keys and credentials. Every shape is now covered by a test that applies the template for real and
inspects the result.

Three documentation corrections were made because the old text had become actively misleading:

- Deleting the settings file was described as safe. It is not, any more. Plugin state is now the one
  managed thing the template reads rather than declares, so deleting the file switches every disabled
  plugin back on and drops any version pin. That advice appeared exactly where someone would reach for it
  during a problem.
- The promise that switching a plugin off sticks is true for the plugins this repository declares, and
  not for any other. That distinction is live today rather than theoretical: skills installed as folders
  are loaded as plugins too, and switching one of those off is still undone by an apply.
- A rule about what happens to a plugin left out of the list was correct but unscoped, and read as a
  general claim about all plugins when it only holds for one kind.

## How it was reviewed

Five rounds, and the reviews changed the design more than once.

The first review rejected the approach outright by building it and running it, rather than reading it.
The security review found that the repository had stopped re-asserting plugin state at all, which is a
real cost that the original write-up had understated. A test-quality review ran fifty-seven deliberate
breakages against the guard and found eight that it failed to notice, every one of them a case where a
setting kept its shape while losing its contents; one of those silently re-enabled deletion of session
history.

One finding was rejected. It arrived with a function extracted from the shipped binary and a coherent
argument, and it was wrong: it described how a plugin's on/off state is resolved, when the question was
how a plugin gets considered at all. It was caught because the fix round was asked to re-verify it rather
than apply it. The rule it would have overturned was left standing, and scoped instead.

## Effect of merging

Nothing changes until settings are next applied. After that, a plugin you switch off stays off, and the
cutover will not switch it back on.

The trade is stated plainly in the documentation rather than left implicit: the repository no longer
re-asserts this part of the settings file, so if something else changes it, no drift check here will
report it. Recovering from switching everything off is one command per plugin, because there is no
bulk re-enable.


webdavis:
### Pipeline gates for PR #133 (Strategy-A)

```
slice-checklist: checklist-t87-plugin-disable.md against Strategy-A (security)

  1     done [filefile] /private/tmp/claude-501/-Users-stephen--he
  2     done [filefile] `VERDICT: FINDINGS (6)` from an independen
  3     done [commitcommit] commits ffe39269 (fix + guard restructure 
  4a-s  done [file] /private/tmp/claude-501/-Users-stephen--he
  4b    done [commitcommit] `VERDICT: FINDINGS (8)` from an independen
  6     done [commitcommitcommitcommitcommitcommit] commits da6d897c, 9a88111a, 4f774141, a4cb
  6v    n/a
  7     done [commitcommit] `VERDICT: FINDINGS (3)` from an independen
  9     done [commit] 54cea43c7a4fefbdeef0bb3f1fd659fef60b9e6c p

findings-register: findings-t87-plugin-disable.md against Strategy-A

  F1   fixed+proven  ffe39269 test/unit/claude-enabled-plugins.sh k
  F2   fixed+proven  ffe39269 test/unit/claude-enabled-plugins.sh r
  F3   fixed (notest) ffe39269 uses the full name@marketplace keys 
  F4   fixed+proven  ffe39269 test/unit/claude-enabled-plugins.sh r
  F5   accepted      Operator ruling: accept it, a disable always s
  F6   fixed (notest) c5db5d1f rewrites that section plus the templ
  F7   fixed+proven  9a88111a and a4cbe040 test/unit/claude-enabled
  F8   fixed (notest) c5db5d1f documents the trade in CLAUDE.md and
  F9   fixed+proven  4f774141 test/unit/claude-enabled-plugins.sh A
  F10  fixed (notest) c5db5d1f and 4a74d6f6 qualify the claim. No t
  F11  fixed+proven  da6d897c test/unit/claude-enabled-plugins.sh g
  F12  fixed (notest) 6507d219 rewrites the recovery in both places
  F13  fixed (notest) 4a74d6f6 scopes the promise in CLAUDE.md, the
  F14  accepted      CONTRADICTED on re-verification and deliberate

    4a    ok            1 finding(s)
    4a-s  ok            3 finding(s)
    4b    ok            1 finding(s)
    7     ok            3 finding(s)

  disposition: 12 fixed, 2 accepted, 0 deferred (of 14)

```

PR #134, opened 2026-08-02
feat(skills): declare the five kepano/obsidian-skills store entries

webdavis:
## Context

This repository keeps a declared list of the agent skills that belong on this machine, so a rebuilt
machine gets the same set and a weekly job keeps them current. Five skills for working with Obsidian
vaults were installed some time ago and never added to that list.

Nothing was broken by the omission, but the five were invisible to the repository: a fresh machine would
not have installed them, and the weekly refresh did not know they existed.

## Summary

- Declares the five Obsidian skills so a rebuilt machine installs them and the weekly refresh maintains
  them.
- Sets them to load only when asked for by name, matching how every other narrow-purpose skill here is
  treated. They load automatically today, so this is a deliberate change and its cost is described below.
- Closes a hole in the guard that checks these declarations: it compared the names of the links it
  creates but never where they point, so a mistyped destination would have been accepted silently.
- Pins the four counts this repository states about its own skill list, which nothing was checking. Three
  of those numbers are edited by this change.

## What changed

The five skills are `defuddle`, `json-canvas`, `obsidian-bases`, `obsidian-cli` and `obsidian-markdown`,
all from the same upstream project. Their source was read from the installer's own records rather than
guessed, and each is declared in the four places the guard requires.

The two guard fixes are the more interesting part, and both were found by deliberately breaking things to
see whether the tests noticed.

The link check compared filenames only. Since this change hand-writes five link destinations, a typo in
any of them would have produced a dead link, and the skill would simply never load, with nothing
reporting a problem. The check now verifies where each link points, and was confirmed to pass unchanged
against the previous state of the repository, so it tightens the rule without excusing anything that came
before.

The counts this repository publishes about itself were not checked at all. Reverting all four to their
old values left every test passing. They are now pinned, with two details that matter: a check that
matches nothing has to fail rather than pass, otherwise rewording a sentence quietly retires its own
check; and the matching tolerates the line wrapping that the formatter applies, since one of the
sentences is already split across lines.

## How it was reviewed

Four rounds. The first, before any code existed, corrected four things: two of the counts in the plan
were wrong, one skill it proposed to add an exemption for was already fully declared, and its instruction
to add a companion file for each skill would have failed the build outright, which the reviewer
demonstrated rather than argued.

The correctness round found nothing, having first confirmed it was reviewing the actual branch by
comparing hashes. The test-quality round ran eighty-two deliberate breakages and found the two gaps
described above, then fixed them.

## Effect of merging

Nothing changes until settings are next applied. After that the five skills stop loading automatically
and load only when named.

One of them, `defuddle`, exists to be used automatically: it reads web pages in place of the built-in
fetch. Once it loads on request only, pasting a link will not reach it unless it is asked for by name.
That is reversible with a one-line change if it turns out to be the wrong call.

Two skills sitting in the store are reported and deliberately untouched: one is genuinely superseded, and
one is still referenced by name from two skills that are declared here, so it is not the leftover it
appears to be. A separate task covers them, along with a wider mismatch between what this repository
declares from that upstream project and what the project still ships.


webdavis:
### Pipeline gates for PR #134 (Strategy-A)

```
slice-checklist: checklist-t93-obsidian-skills.md against Strategy-A

  2     done [file] /private/tmp/claude-501/-Users-stephen--he
  3     done [commitcommitcommitfilefile] commits 4e69cfec (harden the guard), 85f82
  4a-s  n/a
  4b    done [commitcommitfilefile] `VERDICT: FINDINGS (2)` from an independen
  6     done [commit] n/a in effect. Step 4a returned CLEAN so t
  6v    n/a
  7     deviated      DEVIATION, step 7 did not run because there was no FIX to 
  9     done [commit] 1e1fa85ee12c186dbf8d92a9c86c95b9e84e5ca1 p

findings-register: findings-t93-obsidian-skills.md against Strategy-A

  F1   fixed (notest) 85f82e9c carries the corrected counts in CLAU
  F2   fixed (notest) 85f82e9c contains no exemption and does not t
  F3   fixed (notest) 85f82e9c ships no `agents/openai.yaml` and no
  F4   accepted      Corrected in the step-3 dispatch, which instru
  F5   accepted      Deliberately NOT built, per the step-3 dispatc
  F6   fixed+proven  10d75d01 test/unit/skills-roster-fanout.sh rul
  F7   fixed (notest) 0da637d5 corrects the claim and adds a note n

    2     ok            5 finding(s)
    4a    ok            0 finding(s)
    4a-s  n/a
    4b    ok            2 finding(s)
    7     n/a

  disposition: 5 fixed, 2 accepted, 0 deferred (of 7)

```

PR #135, opened 2026-08-03
feat: weekly unattended jobs record what they changed to #unattended-upgrades

webdavis:
## Context

This machine runs two unattended weekly jobs: one refreshes the agent-skills store, the other upgrades
Homebrew packages. Until now they only spoke up on failure, and one of them not even then. A quiet week
and a dead job produced identical silence, and when the machine later misbehaved there was no record of
what had changed to investigate against.

This branch gives both jobs a weekly record, delivered to a dedicated Discord channel, while failures
keep going to the existing alert channel. Record one, act on the other.

## Summary

- Adds an `unattended-upgrades` webhook route to the encrypted hermes config, sharing the existing relay
  route's signing key, and posts one entry per job per week to it.
- Each entry states its own gap: when this run happened, when the last successful run happened, and how
  long that is. A reader of only the newest message can tell a healthy job from a dead one.
- A run that changes nothing still posts, and a run that is deferred posts that it attempted nothing.
  On this machine the deferred entry is the one that will actually appear, which is itself a finding
  (the skills updater has never completed a run here; tracked separately).
- `relay.sh` gains a `--remote-only` mode for these entries: no desktop banner, no phone push, and the
  delivery outcome is reported instead of discarded. A week whose entry cannot be delivered stays
  unclaimed so later slots retry, and one alert goes to the priority channel.
- The Homebrew job, which previously reported nothing at all, now relays its failures too, and a plist
  marker stops a hand-run `just brew-upgrade` from forging the weekly entry.

## What changed

The shared entry logic lives in one library file so both jobs report in the same shape. The week claim
is an exclusive file creation, so 24 scheduled Monday slots produce one entry rather than 24, and the
claim is released if delivery fails, so a broken channel never reads as a delivered week.

What an entry can honestly say differs by lane and the entry says so: Homebrew formulae and App Store
apps have real versions; most skills are installed unpinned from a repository head, so their entries
report content-hash change only and state that no version number is knowable.

Names and version strings from third parties are rendered in code spans with control characters
stripped, so a hostile or merely unlucky upstream string cannot inject Discord markup, forge an extra
line, or inflate the counts.

## How it was reviewed

Five review rounds shaped this: a brief review that reversed two of the design's own premises before
any code existed, a correctness review that found the week guard could mark an undelivered week as done,
a security review that proved the shared-secret decision from the gateway source and caught the markup
injection, a test-quality pass that ran 22 mutations and pinned six behaviors nothing was holding, and
[...]

PR #136, opened 2026-08-03
fix: reconcile the committed hermes config with the live gateway state

webdavis:
## Context

This repository keeps the hermes gateway's configuration as an encrypted file, and the cutover apply
will overwrite the live config with it. The two had drifted apart in eleven places, so the apply would
have silently regressed the running gateway: reverted its default model to an older one, turned off a
toolset in daily use, and dropped settings that were changed live after the file was last captured.

## Summary

- Adopts ten live values into the committed config: the current default model, the kanban toolset in
  both places it is referenced, the ponytail plugin, a command-allowlist entry, auto-threading for
  Discord, the compression threshold, the kanban orchestrator profile, and multiplexing off at both of
  its paths.
- Keeps two divergences on purpose, documented in the commit: the terminal working directory stays
  portable (`.`) rather than adopting a machine-specific absolute path, and the log route added by
  #135 stays committed-only until the cutover deploys it.
- Each doubtful entry was decided by an explicit ruling rather than adopted blindly: the allowlist
  grant and the multiplexing flip were confirmed as intended, the working-directory write-back was not.

## How it was reviewed

The brief review re-measured everything fresh and caught three errors in the original measurement
(values that read as absent were actually present-and-false, a comparison artifact), found a twelfth
divergence the plan had missed, and flagged that the plan's own success criterion would have rewarded
deleting the new log route. The final check decrypted both versions independently and confirmed the
change is exactly the ten intended edits, nothing else, with every route secret byte-identical.

## Effect of merging

Nothing changes on the live machine now. At the cutover apply, the gateway's config file will match
what is actually running today, instead of regressing it, with the two documented exceptions: the
working directory resets to portable form, and the new log route becomes active.


webdavis:
### Pipeline gates for PR #136 (Strategy-A)

```
slice-checklist: checklist-t96-hermes-drift.md against Strategy-A (security)

  2     done [file] VERDICT: FINDINGS (6). MODEL-VERIFIED: cla
  3     done [commitcommitcommitfile] 2 commits, reconciliation 2561061e plus th
  4a    done [filefilefile] VERDICT: CLEAN (`"verdict":"CLEAN"`, `"fin
  4a-s  deviated      DEVIATION: no separate security pass. The slice is a 16-li
  4b    deviated      DEVIATION: no test-quality pass. The slice changes zero te
  4c    done [commit] Independent 12-point verification by the o
  5     done [commit] Nothing to adjudicate: 4a returned CLEAN w
  6     deviated      DEVIATION: no fix round; there are no findings to fix (4a 
  6v    n/a
  7     deviated      DEVIATION: no re-review; there is no fix to re-review. The
  8     done [file] Zero deferrals: all six step-2 findings FI
  9     done [commit] Gates green in the flake shell per steps 3

findings-register: findings-t96-hermes-drift.md against Strategy-A

  F1   fixed (notest) Closed in the step-3 dispatch before any code
  F2   fixed (notest) Same dispatch: fetch-then-branch from origin/
  F3   fixed (notest) Orchestrator reproduced with a type-aware wal
  F4   fixed (notest) Escalated to the OPERATOR, who ruled 2026-08-
  F5   fixed (notest) Dispatch corrects to expect 4-vs-3. Brief-sta
  F6   fixed (notest) Dispatch states PLAINTEXT explicitly. Brief-s

    4a    ok            0 finding(s)
    4a-s  n/a
    4b    n/a
    7     n/a

  disposition: 6 fixed, 0 accepted, 0 deferred (of 6)

```

PR #140, opened 2026-08-04
feat(claude-commands): rebuild the git and pull request slash commands on gh-axi

webdavis:
## Context

Claude Code's `commit-commands` plugin ships convenient one-shot git commands, and it is installed here
but disabled for a concrete reason: its `/commit-push-pr` hard-codes `gh pr create` and allowlists
`Bash(gh pr create:*)`, which conflicts with the standing rule that every GitHub operation goes through
gh-axi (a wrapper CLI invoked as `npx -y gh-axi`) and never through the bare `gh` CLI.

This branch rebuilds the useful commands as custom slash commands in `private_dot_claude/commands/`
(chezmoi's source for `~/.claude/commands/`, which already held `pr-merge.md`), with the rules written
into each prompt rather than left to the executing model's recall of CLAUDE.md. The plugin stays
disabled.

## Summary

- Adds seven command prompts under `private_dot_claude/commands/`: `commit.md`, `commit-push-pr.md`,
  `pr.md`, `clean-gone.md`, `amend.md`, `uncommit.md`, and `sync-worktrees.md`.
- Both pull-request commands (`commit-push-pr.md` and `pr.md`) inline the locked five-section body
  template and a mandatory self-review step that sits above the `gh-axi pr create` call in the file, so
  the body cannot reach GitHub unreviewed. The body is passed with `--body-file`.
- Fixes `private_dot_claude/commands/pr-merge.md`, which described a squash merge and called the bare
  `gh` CLI. It now runs `npx -y gh-axi pr merge <number> --merge --delete-branch` with the subject
  `Merge pull request #N from webdavis/<branch> (#N)`.
- `clean-gone.md` and `sync-worktrees.md` defer to worktrunk instead of fighting it: a gone branch that
  owns a worktree is routed through `wt remove`, and `sync-worktrees` is a thin wrapper of `wt up`.
- Adds `test/unit/claude-commands-govern.sh`, 120 assertions over the eight prompts, each one
  mutation-verified against an unmutated control (table below).

## How it was verified

- `just l` (treefmt over the tree): traversed 603 files, emitted 374, **0 changed**.
- `shellcheck -x test/unit/claude-commands-govern.sh`: clean.
- `just test` (all four suites): exit 0, zero `not ok` and zero failing scripts, 328 bats assertions
  passing.
- `test/unit/claude-commands-govern.sh` standalone: `PASS: 120 assertions over 8 command prompts`.
- Each of the three commits passed the pre-commit gate on its own (unit suite plus a gitleaks scan
  reporting no leaks).

Every assertion was mutation-verified. Each row applies one mutation to a pristine snapshot, runs the
guard, records the first failure, and restores; the last row is the unmutated control.

| # | Mutation | Verdict | First failure reported |
| - | -------- | ------- | ---------------------- |
| M01 | swap `gh-axi` for `gh` in a create call | RED | `pr.md` invokes the bare gh CLI |
| M02 | stop naming gh-axi at all (`pr-merge.md`) | RED | never names gh-axi |
| M03 | allowlist `Bash(gh pr create:*)` in frontmatter | RED | allowlists the bare gh CLI |
| M04 | delete the `## Effect of merging` heading | RED | omits that template section |
| M05 | swap `## Summary` and `## How it was verified` | RED | template order is locked |
| M06 | delete the review step heading | RED | has no mandatory review step |
| M07 | delete the em-dash item from the checklist | RED | checklist no longer mentions it |
| M08 | move the create call above the review step | RED | body can be posted unreviewed |
| M09 | stop passing the body as a file | RED | does not pass the body with `--body-file` |
| M10 | pass the body inline instead | RED | passes the body inline |
| M11 | stop invoking the `conventional-commits` skill | RED | does not invoke the skill |
| M12 | delete the Co-Authored-By prohibition | RED | does not forbid the trailer |
| M13 | drop "per-invocation" from a confirmation | RED | does not demand per-invocation confirmation |
| M14 | delete the "does not carry over" clause | RED | an earlier yes would still count |
| M15 | insert an em-dash | RED | contains an em-dash |
| M16 | reintroduce the squash wording in `pr-merge.md` | RED | still mentions squashing |
| M17 | drop the `--merge` flag | RED | does not pass `--merge` |
| M18 | drop the merge subject convention | RED | does not spell out the subject |
| M19 | add an ungoverned `rogue.md` | RED | roster and directory disagree |
| M20 | delete a command file | RED | roster and directory disagree |
| M21 | empty a command file | RED | missing or empty |
| M22 | drop the frontmatter fence | RED | no `---` fence, so no metadata |
| M23 | drop the frontmatter `description` | RED | has no description |
| M24 | drop every `allowed-tools` entry | RED | allowlist is empty |
| M25 | drop the gh-axi allowlist entry | RED | create call is not permitted |
| M26 | CONTROL, no mutation | GREEN | `PASS: 120 assertions` |

M04 exposed a real weakness on the first pass. The heading-presence assertion used substring matching,
so deleting `## Effect of merging` from the template block still satisfied it: the prose bullet that
explains that section repeats the same words. The guard went red anyway, but through the ordering check
rather than the assertion that should have caught it. The heading assertions now match whole lines, and
M04 fails on the heading itself.

## Effect of merging

Merging changes nothing on a live machine. These are plain files, so they reach `~/.claude/commands/`
only at the next `chezmoi apply`, and no apply has been run (pre-cutover freeze). The commands are inert
until someone invokes one.

It does not re-enable the `commit-commands` plugin, and it does not touch the global CLAUDE.md rule that
routes pull request descriptions through the humanizer skill; that wording update belongs to the
separate global-rules refactor.

One property worth knowing: `treefmt.nix` excludes `private_dot_claude/commands/**` from mdformat, so
that YAML frontmatter survives (mdformat without the frontmatter plugin turns `---` into a horizontal
rule plus a heading and breaks command discovery). The 105-column wrap in these files is therefore held
by hand and by review; the guard checks content rules, not line width.

## Review guide

Read `private_dot_claude/commands/commit-push-pr.md` first. It is the longest prompt and every other
command is a subset of its rules. `pr.md` repeats the same body machinery without the commit and push
steps; that duplication is deliberate, since a slash command cannot include another one.

`test/unit/claude-commands-govern.sh` is the load-bearing file. The assertion worth the closest look is
the review-step ordering, which compares line numbers rather than presence: a prompt is read top to
bottom, so a review instruction below the create call is a review that never runs.

`pr-merge.md` is a small independent fix and sits in its own commit. Everything else in the diff is
prose.

### DIFF
--- a/private_dot_claude/commands/amend.md
+++ b/private_dot_claude/commands/amend.md
@@ -1,0 +1,46 @@
+---
+description: Amend the last commit, refusing to rewrite pushed history without explicit confirmation.
+allowed-tools:
+  - Bash(git status:*)
+  - Bash(git diff:*)
+  - Bash(git add:*)
+  - Bash(git commit:*)
+  - Bash(git log:*)
+  - Bash(git branch:*)
+  - Bash(git rev-parse:*)
+  - Bash(git rev-list:*)
+  - Skill
+  - Read
+---
+
+# Amend
+
+Fold the current changes into the last commit, keeping or regenerating its conventional message.
+
+## Steps
+
+1. Run `git log -1 --format='%h %s'` and `git status --short` so the target commit and the pending
+   changes are both on the table.
+1. Decide whether the commit is published: `git rev-parse --abbrev-ref '@{u}'` names the upstream,
+   and `git rev-list --count '@{u}'..HEAD` reports how many local commits sit above it. A count of
+   zero means HEAD is already on the remote, so amending rewrites published history.
+1. When HEAD is published, stop and say so. Amending it needs a follow-up force-push, so it takes
+   per-invocation confirmation from the operator. A yes given earlier does not carry over. Wait for
+   an explicit answer before touching anything.
+1. Stage the changes that belong in the commit, by path.
+1. Amend. Keep the existing message with `git commit --amend --no-edit` when the subject still
+   describes the commit. When it no longer does, invoke the `conventional-commits` skill, write a
+   fresh message, and pass it with `git commit --amend -m "<subject>"`.
+1. Report the rewritten commit's short SHA and subject, and, when it was published, the exact
+   force-push command the operator now needs.
+
+## Safeguards
+
+- The message never carries a `Co-Authored-By` trailer, any Claude or Anthropic co-author line, or a
+  "Generated with Claude Code" footer.
+- No em-dashes in the message.
+- Never amend a commit on `main`.
+- Never run the force-push yourself as part of this command. When one is needed it is
+  `git push --force-with-lease`, never `--force`, and it is the operator's call.
+- Bail when HEAD is a merge commit, when a rebase or cherry-pick is in progress, or when the
+  repository has no commits yet. Report which one blocked it.
--- a/private_dot_claude/commands/clean-gone.md
+++ b/private_dot_claude/commands/clean-gone.md
@@ -1,0 +1,44 @@
+---
+description: Prune local branches whose upstream is gone, worktrunk-aware, deleting nothing unconfirmed.
+allowed-tools:
+  - Bash(git fetch:*)
+  - Bash(git branch:*)
+  - Bash(git worktree:*)
+  - Bash(git status:*)
+  - Bash(git rev-parse:*)
+  - Bash(wt:*)
+---
+
+# Clean Gone Branches
+
+Delete local branches whose remote tracking branch is gone, without fighting worktrunk.
+
+Worktrunk owns the worktrees in this repository and sets `delete-branch = false`, so a branch ref
+that outlives its merged pull request is deliberate, not leftover garbage. Treat that config as
+correct and never change it from here.
+
+## Steps
+
+1. Run `git fetch --all --prune`.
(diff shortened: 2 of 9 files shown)


webdavis:
Reviewed all nine files against the slice spec.

Verified: gh-axi everywhere with no bare `gh` token or allowlist entry; the five-section body template inlined verbatim in both PR-creating commands with the mandatory review step ordered before the create call and `--body-file` enforced; per-invocation confirmation language on every destructive path (clean-gone deletions, amend/uncommit on published commits, force-push); worktrunk ownership respected in clean-gone and sync-worktrees; pr-merge now carries the `--merge` + subject convention. Tool claims spot-checked against the real CLIs: `wt remove --force-delete` exists and `wt up`'s alias body matches the description in sync-worktrees.md line for line. The mutation table (25 RED, control GREEN) held up, and the M04 substring fix is the right call.

Two cosmetic notes, no change requested: the `/tmp/pr-body.md` example path (an executing session substitutes its own scratch location), and pr-merge.md keeping its original frontmatter-less shape.

Approved.

PR #141, opened 2026-08-04
feat(claude): declare plugins and marketplaces, and record what auto-update changed

webdavis:
## Context

This is the chezmoi dotfiles repository, where `private_dot_claude/modify_settings.json` is a
modify-template that enforces a fixed set of fields in `~/.claude/settings.json` while letting `/config`
toggles drift freely, and `~/.local/bin/update-skills.sh` is the weekly job that keeps the canonical
skills store (`~/.agents/skills`) fanned out to Claude Code, Codex and hermes.

An earlier version of this work built a 981-line weekly updater that installed and upgraded Claude Code
plugins on a timer. Claude Code already does that itself at startup, so the updater and its lock file are
dropped. What survives is the part Claude Code does not do: declaring which plugins and marketplaces this
machine wants, and keeping a record of what its silent auto-updates changed.

## Summary

- The settings modify-template declares twelve plugin ids instead of nine, adding
  `skill-creator@claude-plugins-official`, `worktrunk@worktrunk` and `last30days@last30days-skill`.
- It also declares the four marketplaces Anthropic does not publish (`ponytail`, `openai-codex`,
  `worktrunk`, `last30days-skill`) with `autoUpdate` set to `true`, one `setValueAtPath` per marketplace
  key so a marketplace added with `claude plugin marketplace add` keeps its own entry.
- `dot_agents/custom-skill-lock.json` gains a `claudeDelivery` table, and `update-skills.sh` subtracts
  its `"none"` set from the weekly Claude fan-out. `last30days` is its one entry: the store copy still
  serves Codex and hermes, and the `~/.claude/skills` symlink declaration is gone.
- New `dot_local/bin/executable_report-plugin-updates.sh`: a read-only weekly job that compares
  `~/.claude/plugins/installed_plugins.json` against a snapshot under `~/.local/state/` and posts the
  plugin-by-plugin version changes to the `#unattended-upgrades` channel through `relay.sh`.
- New `Library/LaunchAgents/com.webdavis.report-plugin-updates.plist.tmpl` (Monday 13:00,
  `RunAtLoad=false`) plus its `.chezmoiscripts/run_onchange_after_69-` loader.

## What changed

**Marketplace auto-update.** Read out of the shipped Claude Code 2.1.220 binary: the settings schema
declares `autoUpdate` as an optional boolean per `extraKnownMarketplaces` entry, and an entry that omits
it falls back to membership of a built-in set of Anthropic marketplace names. So the four third-party
marketplaces on this machine defaulted to never updating, and nothing said so. The same schema declares
`source` required, which is why each entry is written whole rather than as a lone flag.

**The `claudeDelivery` table.** The Claude fan-out used to link every store entry unconditionally, so a
`~/.claude/skills` link removed by hand came back on the next Monday. A key mapped to `"none"` is now
skipped by the fan-out and carries no chezmoi declaration. The table states only what the skills vertical
does; it names no other delivery mechanism and reads no other lock. A malformed table refuses the run
rather than failing open, in every mode including `--dry-run`, which reaches the fan-out without the
roster snapshot gate.

**The reporter.** Its source of truth is `~/.claude/plugins/installed_plugins.json` (schema version 2,
checked against the live file). `known_marketplaces.json` and `plugin-catalog-cache.json` were both
inspected and rejected: one records marketplaces rather than versions, the other lists what is available
rather than what is installed. The fingerprint is `version`, falling back to `gitCommitSha`, falling back
to the literal `unknown`; `lastUpdated` was rejected as a further fallback because six plugins carried
their marketplace's own `lastUpdated` to the second, so a plain marketplace refresh would have reported
all six as changed every week. Only plugin ids and fingerprints reach the channel, never an `installPath`
or a marketplace URL. The entry shape is `dot_local/bin/unattended-log-lib.sh`, shared with the weekly
Homebrew upgrade and the weekly skills update.

The rules it holds: a first run records a baseline and posts nothing. A quiet week still posts and names
zero changes, because otherwise a clean week and a dead LaunchAgent produce identical silence, which is
the reason `#unattended-upgrades` exists at all. An unreadable state file posts no record and alerts on
the priority route instead, since the only change list it could build from a file it cannot read is a
false "nothing changed". The snapshot moves only after the gateway accepts an entry, so a refused
delivery is reported by the next run rather than lost.

The one deliberate deviation from the brief: the brief said a no-change week should post nothing, with a
parenthetical allowing the established convention instead. The convention posts, and posting is what the
shared library was built for, so a quiet week gets an entry naming zero changes. The tests pin both
halves, that the quiet entry never fabricates a change and that a real change is never silent.

## How it was verified

`just l` clean. All four suites green on this branch's tip: unit, integration, e2e and test-system each
exited 0. Every new assertion was checked by mutation against an unmutated control:

- Settings template: flipping a declared `autoUpdate` to `false` and replacing the four per-key writes
  with one whole-map write both fail `test/unit/claude-enabled-plugins.sh`, the second on the new
  operator-added-marketplace case.
- Fan-out: deleting the `claudeDelivery` subtraction fails `test/integration/update-skills-converge.sh`;
  repointing the lock read at another table and writing an illegal table value both fail
  `test/unit/skills-roster-fanout.sh`.
- Reporter: freezing the fingerprint, deleting the empty-inventory refusal, moving the snapshot before
  the post, and dropping the `--scheduled` gate all fail `test/e2e/report-plugin-updates-record.sh`. The
  empty-inventory mutation survived the first version of that test, because `jq -e` exits non-zero on an
  empty result and masked it; the assertion now requires the alert to carry a reason.
- Plist: giving the agent the same slot as the Homebrew job, and pointing the loader at a different log
  directory, both fail `test/unit/report-plugin-updates-plist.sh`.

The reporter was also run by hand against this machine's real
`~/.claude/plugins/installed_plugins.json` with its state directory pointed at a scratch path: it read 17
user-scope plugins, and against a seeded snapshot it rendered a real
`exa@claude-plugins-official 3.3.0 -> 3.4.0` transition while leaving an unchanged plugin out of the
list. Being a manual run it posted nothing and moved nothing, which is the gate step 7 pins.

## Effect of merging

Nothing on the live machine changes on merge. Every part of this lands at the next `chezmoi apply`, which
is the D1 cutover.

After that apply: the three added plugin ids are enabled and the four marketplaces begin auto-updating at
Claude Code startup, and `com.webdavis.report-plugin-updates` is loaded and starts its Monday cadence,
recording a baseline on its first run and posting from the second. A live `claude plugin disable` still
survives an apply for all twelve ids, unchanged by this work.

The apply does NOT remove the existing `~/.claude/skills/last30days` symlink; chezmoi leaves a target
behind when its source declaration goes away, and no `.chezmoiremove` is added here. The next weekly
`update-skills.sh` run reaps it, because that link is updater-owned and no longer in the fan-out's desired
set. Removing it by hand is the same outcome a week earlier.

## Review guide

Start with `private_dot_claude/modify_settings.json`, where the marketplace block sits below the plugin
roster and both carry their reasoning inline. Then `dot_local/bin/executable_update-skills.sh`, three
hunks: the schema gate, the two `__update_skills_claude_*` helpers, and the subtraction inside
`converge_claude_skills`. The reporter is one file,
`dot_local/bin/executable_report-plugin-updates.sh`, and reads top to bottom.

The assertion worth the most attention is step 3 of `test/e2e/report-plugin-updates-record.sh`: a
reporter that stays silent while a plugin really moved is the failure this whole change exists to
prevent, and that case is what pins it.


webdavis:
Reviewed the declarations, the update-skills hunks, the reporter, and the tests.

Verified: the three roster additions are per-key setValueAtPath writes so /config-owned state keeps drifting freely, and the four extraKnownMarketplaces entries fix a real defect confirmed at the binary level (unset autoUpdate falls back to an Anthropic-marketplaces allowlist, so ponytail, worktrunk, last30days and codex would never update). The reporter posts plugin names and versions only, gates posting and snapshot movement on --scheduled, fails loud on a malformed inventory, and moves the snapshot only after a successful post; the mutation table demonstrates each of those properties can go red, including the jq -e masking hole the first version had. The symlink_last30days removal is the right dual-management split: Claude takes the plugin, the store copy keeps serving Codex and the hermes concerned profile, and the live-symlink-survives-until-reaped caveat is stated rather than hidden. Hunks against the roster test do not overlap #138's (mergeable: true against current main).

All four logged deviations are accepted with their reasons, the quiet-week entry in particular: a clean week and a dead LaunchAgent must not produce identical silence.

Approved.

PR #142, opened 2026-08-04
fix(ssh-hardening): bound the child verify so a hang rolls the install back

webdavis:
## Context

This is the chezmoi source tree for the dotfiles. One of the scripts it manages,
`dot_local/bin/executable_ssh-hardening.sh`, installs a public-key-only sshd drop-in on macOS. Installing
is written as a transaction: stage the new policy file, publish it with one rename, move the legacy
drop-in aside, verify the configuration sshd actually resolves, and put the tree back exactly as it was
found if that verify reports a failure.

The transaction held only against a verify that RETURNS. A verify that hangs returns nothing to roll back
on, so the run parked with the new drop-in published and the legacy file gone, and the rollback was never
reached. That is measured rather than theoretical: a single named pipe in the drop-in directory blocks
`sshd -G` forever, because sshd resolves its own Include globs with no type filter, and the install's
verify runs sshd.

## Summary

- The child verify now runs under a deadline in `dot_local/bin/executable_ssh-hardening.sh`
  (`VERIFY_DEADLINE_SECONDS`, 120 seconds, with `SSH_HARDENING_VERIFY_DEADLINE` overriding it for tests).
  It is a poll around a background job, not `timeout(1)`, which stock macOS does not ship.
- An expiry returns nonzero, which is what both callers of `run_verify_child` already read as a failed
  verify: the install takes its existing rollback, and `--reload` refuses to restart. No rollback logic
  was added or changed, and a loud `verify TIMED OUT` line says which path fired.
- The stopped verify is signalled as a process group (TERM, a 2 second grace, then KILL), so the sshd
  blocked underneath the child shell is stopped with it instead of orphaned.
- `test/e2e/ssh-hardening-verify-watchdog.sh` pins three outcomes: a wedged verify, a healthy verify, and
  a verify that fails by answering. The hang is built with the ssh fixtures' existing
  `SSH_TREE_MUTATION_HOOK` seam, which runs inside the sshd stub.
- The install-section comment that argued against a wall clock is replaced by the measurements and the
  cost argument that answer it: a rollback restores what the run found, so an expiry costs an install
  that did not apply, not the hardening of a machine that was fine.

## How it was verified

Red for the right reason. The new e2e file, run with `SSH_HARDENING_SCRIPT` pointed at the pre-fix script
from `HEAD`, fails on the harness clock rather than hanging the suite:

```
FAIL: run_ssh_reload  exceeded its 30s wall clock; a reload that can hang is itself a regression
```

The same file with the wedged case cut off passes against that same pre-fix script, so the
healthy-verify and answered-failure cases are regression pins rather than restatements of the new
behavior:

```
pins-only (cases 3 and 4) PASSED
```

Green with the fix, in 5 seconds (a 3 second test deadline plus the 2 second stop grace):

```
ssh-hardening-verify-watchdog: OK (a verify wedged inside sshd is stopped at its deadline in 5s,
reported as a failed verify, rolled back to a byte-identical tree with no working files and no
surviving processes; a healthy verify still installs with exactly its three seam calls; a verify
that answers with a failure still rolls back and is not reported as a timeout)
```

No orphans. The wedge records the sshd stub's pid and the pid of the sleeper it becomes, and the test
polls both after the run and fails naming whichever is still alive (polled, because a killed process is a
zombie until it is reaped and `kill -0` answers 0 for a zombie). The mechanism was checked on its own
against bash 3.2 first: `set -m` puts the background job in its own process group, and a group TERM
reaps a grandchild that a kill of the child alone leaves running.

The bound is set from measurement, taken against a sandbox tree driven by the real `/usr/sbin/sshd`
through the fixtures' seams: a three-file tree verifies in 0.22s, and a 1.0 MB tree, the shape
`MAX_CONFIG_TREE_BYTES` was set against, verifies in 5.1s.

Every existing ssh-hardening suite passes (the unit and integration files both), then `just l` reported
0 changed and the full `just test` was green.

## Effect of merging

Merging changes the source tree and nothing on this machine. The script reaches
`~/.local/bin/ssh-hardening.sh` at the next `chezmoi apply`, and no apply runs before the D1 cutover, so
the live `/etc/ssh` tree and the running sshd are untouched by this. For an operator running the script
after that apply, behavior is identical except when a verify does not answer inside 120 seconds, where an
install that used to park now rolls back and says so.

## Review guide

- The watchdog block, `run_verify_child` and `stop_verify_child`. What carries weight there: the group
  kill works only because the job is launched under `set -m`, the child is reaped after both kills so its
  pid cannot be recycled underneath them, and the poll's delay tool is deliberately not the `SLEEP_BIN`
  seam, which the suites stub with an instant-return spy.
- The rollback junction, `install_dropin`'s `run_verify_child || verify_status=$?`, which this change
  leaves alone. The question worth asking is whether an expiry can reach any path other than that one.
- The bound itself. It is safe if no tree that would eventually verify can cross 120 seconds; the
  measurements above are the argument, and a counterexample is the useful kind of pushback.
- The narrowing recorded in `run_verify_child`: a child in its own process group no longer dies with a
  Ctrl-C aimed at this shell, so interrupting a hung install inside the deadline leaves that verify
  running where it used to be killed with the shell.

### DIFF
--- a/dot_local/bin/executable_ssh-hardening.sh
+++ b/dot_local/bin/executable_ssh-hardening.sh
@@ -54,6 +54,13 @@
 #                       overridden KEYSCAN_BIN that never returns holds the
 #                       reload for as long as it blocks -- the bound limits
 #                       how many times a probe runs, not how long one takes.
+#   SSH_HARDENING_VERIFY_DEADLINE
+#                       seconds the child verify may run before it is stopped
+#                       and reported as a FAILED verify (default 120; see
+#                       VERIFY_DEADLINE_SECONDS for why that bound is safe and
+#                       what an expiry costs). Anything but a positive whole
+#                       number is the default. The override exists for tests,
+#                       which cannot wait out the real bound.
 #   SSH_HARDENING_ALLOW_MISSING_SSHD
 #                       explicit test seam: when set to a TRUE-ish value AND
 #                       $SSHD_BIN cannot run, --verify skips (exit 0) WITHOUT a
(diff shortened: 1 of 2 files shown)


webdavis:
Reviewed the watchdog, the replaced rationale, and the e2e coverage.

Verified in the code: the deadline constant is validated (non-numeric or zero falls back to 120s rather than becoming a shell error or an instant kill), the poll counts ticks so the fractional interval cannot drift from the seconds the operator message quotes, the child runs in its own process group so the TERM/grace/KILL sequence reaches the sshd underneath the child shell, and stdin comes from /dev/null so a monitored background child cannot be stopped by SIGTTIN. The Ctrl-C narrowing that the process group introduces is paid for in a comment where the cost lives, not hidden.

On overriding the script's prior comment against a verify wall clock: the old argument rested on a wrong premise. rollback_install restores what the run found, so a spurious expiry costs an un-applied install and never removes hardening a machine already had. The measured bound (0.22s typical, 5.1s on a 1.0 MB tree, 120s default) makes spurious expiry remote, and the hang it fixes left the tree half-mutated forever. Replacing the comment with the measurements was the right call.

Evidence accepted: true RED against the unfixed HEAD script through the seam, two regression pins that pass pre-fix (honest pins, not restatements), byte-identical rollback via cksum, zombie-aware orphan polling, and the e2e placement matches the timing-bound rule.

Approved.

PR #146, opened 2026-08-04
feat(osquery): make a file-integrity page triageable

webdavis:
## Context

This is a chezmoi dotfiles repository whose tests run in four camps (unit, integration, e2e,
test-system). Continuous integration is the only gate that runs all four: the commit hook runs the unit
camp alone, and the pre-push hook runs lint only, so a GitHub Actions job on `macos-latest` is the single
place the whole suite executes. CI's verdict is therefore the one signal worth trusting, and anything
that lets it turn on the runner instead of on the code corrupts it.

The osquery security pipeline (a local osquery installation whose findings are routed to pages, digests,
or log-only records) owns roughly sixty test files, the largest concentration in the repository. This
branch audits those files for couplings to the machine and removes the ones that can flip the verdict.

## Summary

- Six suites treated a missing tool as a reason to exit 0. `test/unit/osquery-queue-counts.sh`,
  `test/unit/osquery-store-quote-safety.sh`, `test/unit/osquery-local-notify-durable.sh`,
  `test/unit/osquery-loud-local-status.sh`, and `test/integration/osquery-queue-counts-wal.sh` skipped
  themselves without `sqlite3`; `test/integration/osquery-page-allowlist-seed.sh` skipped without `jq`.
  Each is a hard failure now. Neither tool is a `flake.nix` buildInput, so CI resolves `jq` from the
  runner image and `sqlite3` from `/usr/bin`, and an image change would have turned all six silently
  green.
- `test/e2e/osquery-alerter-hostile-columns.bats` never set `OSQUERY_PIPELINE_MANIFEST`, so its verdict
  fell back to `/var/osquery/pipeline-known-good.sha256`, a real machine path the suite never wrote. It
  now binds a manifest inside its sandbox home, and a new control test asserts that the genuine
  allowlisted agent is suppressed.
- `test/integration/osquery-pipeline-manifest-agreement.sh` raced a backgrounded one-second writer
  against a four-second settle ceiling, and its elapsed-time assertions left five seconds of headroom.
  Each ceiling there is a bound the passing path returns well inside, so all six are wide enough now that
  runner load cannot decide the result.
- Three more waits were too tight for a shared runner: `test/integration/osquery-queue-counts-wal.sh`
  blocked forever on a `read` from a WAL holder that stayed alive without confirming, and it is bounded
  now; `test/e2e/osquery-alert-drainer.bats` gave a forked stub two seconds to write its pid file; and
  `test/integration/osquery-allowlist.bats` gave a lingering grandchild five seconds to outlive a real
  `chezmoi apply`.
- The task named seven tests and ten are changed here, across four coupling classes. Every one is
  reproduced by a mutation in the table below, so none is padding. Three further findings were
  investigated and left alone, with the reasons under the review guide.

## How it was verified

Every reworked test keeps a demonstrated RED path. Each row below was produced by mutating the named
source, running the test, and restoring the source. The control (unmutated) run of each test is green.

| Test | Mutation applied | Result |
| --- | --- | --- |
| `test/unit/osquery-queue-counts.sh` | `sqlite3` removed from PATH | RED, names the missing prerequisite |
| `test/unit/osquery-queue-counts.sh` | corrupt-store branch reports a false zero | RED, `a false zero hides a real backlog` |
| `test/unit/osquery-store-quote-safety.sh` | `sqlite3` removed from PATH | RED |
| `test/unit/osquery-store-quote-safety.sh` | URL quote doubling dropped from the store helper | RED, `URL did not round-trip intact` |
| `test/unit/osquery-local-notify-durable.sh` | `sqlite3` removed from PATH | RED |
| `test/unit/osquery-local-notify-durable.sh` | confirmed banner no longer deletes its row | RED, `expected 0 row(s) once the watcher settled, still 1` |
| `test/unit/osquery-loud-local-status.sh` | `sqlite3` removed from PATH | RED |
| `test/unit/osquery-loud-local-status.sh` | `_loud_local` always returns 0 | RED, `expected a NONZERO status, got 0` |
| `test/integration/osquery-queue-counts-wal.sh` | `sqlite3` removed from PATH | RED |
| `test/integration/osquery-queue-counts-wal.sh` | counters opened `immutable=1`, skipping the WAL | RED, `expected '3', got '0'` |
| `test/integration/osquery-queue-counts-wal.sh` | holder never emits its marker | RED at 5s with the bound; without it, still blocked when an external kill landed at 10s |
| `test/integration/osquery-page-allowlist-seed.sh` | `jq` removed from PATH | RED |
| `test/integration/osquery-page-allowlist-seed.sh` | digest agent's allowlist tuple deleted | RED, names the agent that would self-page |
| `test/e2e/osquery-alerter-hostile-columns.bats` | manifest vouching disabled (the pre-change world) | control RED while the three injection pins stayed GREEN, which is the evidence they were vacuous before |
| `test/e2e/osquery-alerter-hostile-columns.bats` | in-band 0x1F re-split reintroduced into the verdict | `HOSTILE-0x1F` RED |
| `test/integration/osquery-pipeline-manifest-agreement.sh` | settle window never waits | RED, `must resolve to SILENT` |
| `test/integration/osquery-pipeline-manifest-agreement.sh` | settle budget becomes per-finding | RED, `10 misses took 35s` against the 20s ceiling |
| `test/integration/osquery-pipeline-manifest-agreement.sh` | settle wait multiplied 30x (the unbounded shape) | RED, `60s for a 2s window` |
| `test/e2e/osquery-alert-drainer.bats` | banner child inherits the drain lock fd | `T-DRAIN-lock-fd-leak` RED |
| `test/integration/osquery-allowlist.bats` | osqueryi capture inherits the writer lock fd | `the writer's lock never leaks to a child` RED |

The PATH mutations run each test on a mirror of the real PATH with one tool removed, so the guard under
test is the only thing that changes.

The full gate is green. `just l` reformatted nothing, and `just test` exited 0 across all four camps: 201
executable `.sh` tests plus 329 bats tests (264 integration, 46 e2e, 19 test-system), with zero `not ok`
lines.

## Effect of merging

CI going red starts meaning the code is broken again for these ten files. Nothing about the live machine
changes: every file touched is under `test/`, which `.chezmoiignore` keeps out of the target state, so no
`chezmoi apply` behaves differently and no osquery agent, plist, or library is modified. Production
sources were temporarily mutated to produce the table above, and every one was restored from git.

## Review guide

Read in this order.

`test/e2e/osquery-alerter-hostile-columns.bats` carried the worst coupling, and it is the only file whose
assertions changed instead of its plumbing. Its three injection pins were passing while the suppression
path they exist to defeat could not run at all, because nothing in the sandbox could vouch for the
allowlist. The setup block and the new `HOSTILE-control` test are what make the other three mean
something, and this file's two mutation rows are the argument for that.

`test/integration/osquery-pipeline-manifest-agreement.sh` next: six wall-clock sites, with the reasoning
for each widened number in the comment beside it. Check whether a widened ceiling still separates the bug
it pins from a healthy run. For the ten-miss case the bug costs thirty seconds against a twenty-second
ceiling, and the mutation row measuring thirty-five seconds is that check.

The six skip-to-fail conversions are mechanical and read together. The remaining three files are one
changed number or one added flag each.

Three findings were deliberately left alone. The `chezmoi` skip guards in the launch-agent and config
suites cannot fire in CI, because `chezmoi` is a `flake.nix` buildInput and resolves from the nix store;
that is the distinction which put `jq` and `sqlite3` in scope and left these out. The `lockf` skips
assert a darwin-only kernel guarantee, and asserting the host is the shape
`test/unit/no-empty-render-skip-on-darwin.sh` already blesses, as against inferring it.
`test/integration/osquery-feature-set-location.sh` reads through `git grep`, which searches the working
tree and not the index (checked against an unstaged probe line), and it fails closed on any git error.

### DIFF
--- a/dot_local/bin/executable_homebrew-weekly-upgrade.sh
+++ b/dot_local/bin/executable_homebrew-weekly-upgrade.sh
@@ -34,6 +34,14 @@ LOCKFILE="${HOMEBREW_WEEKLY_LOCKFILE:-$HOME/.local/state/homebrew-weekly-upgrade
 STATE_DIR="${HOMEBREW_WEEKLY_STATE_DIR:-$HOME/.local/state/homebrew-weekly-upgrade}"
 LOG_SUCCESS_MARKER="$STATE_DIR/last-success-at"
 LOG_WEEK_GUARD="$STATE_DIR/log-week-claims"
+# The durable record of what THIS run moved, for the osquery file-integrity page
+# to correlate against days later. Keep this literal in sync with
+# OSQUERY_UPGRADE_RECORD in
+# ~/.local/libexec/osquery/results-alerter/file-integrity-triage.sh (the
+# consumer); test/unit/osquery-file-integrity-triage.sh pins them equal, because
+# a rename in one alone leaves that page answering no-record forever, which reads
+# exactly like a quiet month of upgrades.
+UPGRADE_RECORD="$STATE_DIR/last-upgrade-changes.tsv"
 
 # An unknown argument is an ERROR, never a silent fallthrough: a typo'd marker in
 # the plist would otherwise run every week and quietly post nothing, which looks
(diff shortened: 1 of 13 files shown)


webdavis:
Reviewed the four commits end to end: the shared diff extraction, the triage helper, the producer record, and the router attachment.

Verified in code: triage facts attach only after pipeline_verdict has already set the tier, are display-only by construction, and every attachment step is an if-test so a missing helper, a failing helper, or non-JSON output leaves the finding exactly as the verdict produced it under set -e. The mutation table covers the properties that matter: window too wide, window too narrow, record never written, and the guard removed (page lost, exit 3), each red with a green control. Every string added to the render path is apostrophe-free, per the standing alerter constraint.

The design honesty is the strength here: the assumed upgrade record did not exist, so the producer half was built rather than pretended; the correlation is a basename match declared as a lead and never a verdict; the operator-writable state dir is named as a trust limit; and the scheduled-audit page path was left alone with its no-paths decision cited. The sandbox fix for the two harnesses that were writing into the live state dir closed a real hazard this slice would otherwise have shipped.

Approved.

PR #149, opened 2026-08-04
feat(cutover): build the D1 gate runner and the tracked live reconciliation tool

webdavis:
## Context

This repository is being modernized under the SP2 plan
(`docs/superpowers/plans/2026-07-03-sp2-combine-and-split.md`). Its final phase is D1, the cutover:
dresden's live chezmoi source moves off the frozen `integration/modernization` branch onto `main`, in
five sequential gates, after which the three reference pull requests are closed.

The 2026-07-10 audit (matrix row 27) ruled that no cutover command may live as prose inside the plan.
Review rounds 3 through 6 kept finding scoping and fail-open defects in embedded shell that nothing
could hold to code standards, so Phase D now states invariants and pass criteria only, and every command
belongs to a tracked, tested runner merged before S12. That runner was never built, which leaves D1 with
nothing to run. This PR builds it.

## Summary

- `scripts/cutover-gate.sh <1|2|3|4|5>` is the gate runner: one entry point per cutover gate, each
  printing what it will do, running its checks, and ending in a pass, a refusal that names its reason,
  or an operator checkpoint. It implements the plan's 19-item binding acceptance checklist.
- `scripts/live-reconcile.sh` is the tracked reconciliation tool gate 3 runs by absolute path, dry-run
  before live. It converges the declarative skills fan-out to the committed lock (Claude symlinks,
  hermes profile symlinks in both directions, the Codex on-demand overlay, the superpowers routing
  re-assert) and refuses divergence that the update-skills lane or chezmoi owns.
- Gate 1 also refuses when the rendered data view disagrees with the tracked data file. chezmoi's
  data-only reads walk the source directory's nested worktrees ignoring `.chezmoiignore`
  (twpayne/chezmoi#4940), and those subtrees are gitignored, so a clean-tree check cannot see them.
  Measured on dresden: the source declares 54 casks while `chezmoi data` reported 47.
- Gate state lives in `$HOME/.local/state/cutover`, outside the checkout, because gate 1 requires a
  fully-visible-clean tree: the two pins, the expected-delta ledger, the retirement manifest, and the
  per-gate pass markers.
- Tests: `test/unit/cutover-gate-usage.sh`, `test/integration/cutover-gate-preflight.sh`,
  `test/integration/cutover-gate-activation.sh`, `test/integration/cutover-gate-closure.sh`, and
  `test/integration/live-reconcile.sh`, sharing the sandbox builder and boundary stubs in
  `test/integration/helpers/cutover-sandbox.sh`. Nothing here is deployed: `scripts/` and `test/` are
  both listed in `.chezmoiignore`, and no gate runs on merge.

## How it was verified

Test-first throughout: each behavior got a red test before the code that satisfies it. The five suites
cover all 19 checklist items, with the sandbox providing a real git repository at the runner's absolute
repo handle under a scratch `$HOME` and PATH stubs for `launchctl`, `chezmoi`, `just`, `gh`, `tailscale`
and `hermes`. No test touches a live launchd domain, the live checkout, or a real apply.

Mutation testing against an unmutated control, across three rounds: 17 mutants on the original build and
24 more covering every review fix below. All 41 killed, zero survivors. Two mutants survived their first
round and both revealed a weak test rather than a weak fix: one only proved the reconcile tool was
checked somewhere rather than before each invocation, and one built a rename fixture that never produced
a rename. Both tests were rewritten and the mutants then died.

CI caught a portability bug the local suite could not: the soak clock read a file mtime with BSD `stat`
flags first, and GNU coreutils `stat -f` is `--file-system`, which fails on the format operand while
still printing a human-readable block for the file. The runner read that block as an epoch. macOS ships
BSD `stat` and the nix `run` shell CI uses ships GNU, so no BSD host could reproduce it. `mtime` now
tries GNU's `-c` first, BSD's `-f` second, and refuses anything that is not digits; gate 4 is driven
through a stub of each flavor.

The retirement universe derivation was cross-checked against the real repository history: it reproduces
every label the audit's checklist item 13 names, including the out-of-prefix
`com.github.openclaw-setup.watchdog` deleted with no unload, the pre-rename osquery labels, and the
system-domain `systems.nixos.nix-installer.nix-hook` that only a script renders.

### Independent review: 24 findings, per-finding disposition

Every finding was reproduced against the actual code before being fixed. All 24 were real; one was real
in consequence but wrong in mechanism, noted below.

| # | Finding | Disposition | Evidence |
|---|---|---|---|
| :110 | A clean checkout can never pass gate 1 | FIXED | `graphify-out/graph.json` confirmed tracked, so the directory exists in every checkout. The check now targets ignored residue. Case H |
| :107 | Ignored chezmoi sources bypass the clean-tree gate | FIXED | `paseo.json` confirmed present, in `.git/info/exclude`, and listed by `chezmoi managed`. `chezmoi managed` is now diffed against `git ls-files`. Case I |
| :279 | Installer discovery treats test fixtures as real services | FIXED | Reproduced: the scan emitted 4 phantom system services from the test helper, two of them labels that must be RETIRED, so it was cancelling their retirement. Scan limited to deployed paths. Case J |
| :34 | The production base override can erase the expected delta | FIXED | Refused wherever the recorded Phase A base is reachable, which distinguishes the real repository from a sandbox with no marker to forge. Case K |
| :518 | A failed gate 1 restart leaves old passes attached to new pins | FIXED | Confirmed; this also contradicted my earlier deviation note, which only covered the successful path. Invalidation moved to gate entry. Case L |
| :192 | Existence-only pass markers survive failed retries | FIXED | `begin_gate` clears the gate's own marker and everything downstream at entry. Closure case I |
| :552 | The approved retirement manifest remains mutable and unlocked | FIXED | Re-validated against the ledger before execution and again at verification; approval copies the just-derived list, not the proposal it compared. Activation case O |
| :586 | Pins can move during the manual apply checkpoint | FIXED | `--post-apply` now fetches and re-verifies both pins. Activation case L |
| :590 | A failed or skipped staged apply can be recorded as successful | FIXED | Convergence over every managed entry is now required as positive evidence before anything is retired. Activation case M |
| :615 | Rendered data is not rechecked after switching branches | FIXED | The landing stage rechecks it post-checkout. Activation case N |
| :635 | The live reconcile can differ from the dry-run executable | FIXED | Verified against its pinned blob before BOTH invocations; the test edits the tool between them. Closure case J |
| :354 | One path classification approves every hunk in the file | FIXED | Rows are bound to the exact tree-entry pair, so either side changing withdraws the classification. Case M |
| :370 | Rename detection hides the source-side deletion | FIXED | `--no-renames` keeps both sides in the manifest. Case N |
| :348 | Blob equality ignores executable and file-type changes | FIXED | Comparison moved to tree entries, which carry the mode. Case N |
| :348 | Quoted paths can become false landed-unchanged results | FIXED | `-z` output, NUL-delimited read, and two empty lookups now refuse rather than compare equal. Case N |
| :370 | A failed manifest enumerator reads as an empty successful ledger | FIXED | The path list, the plist list and both history walks are materialized through checked commands |
| :447 | Retirement membership discards the launchd domain | FIXED | Membership is by (label, domain) pair. Case O |
| :538 | Every launchctl error is treated as confirmed absence | FIXED | Tri-state probe; verified on this host that not-found is 113, a malformed domain 64, an unreachable GUI domain 112. Activation case P |
| :652 | Historical services loaded after approval are never rejected | FIXED | Gate 3 recomputes the managed extras from the live domains. Closure case K |
| :709 | Drift verification excludes every templated target | FIXED | `--exclude=templates` dropped every templated target, not the KeePassXC-gated ones. Full status now. Closure case L |
| :775 | The end-of-soak gate does not reverify launchd topology | FIXED | Gate 4 re-runs the manifest verification. Closure case N |
| :704 | Heartbeat success does not prove a healthy heartbeat | FIXED | Confirmed every branch ends in `send_alert ... \|\| true`. The gate now asserts canary freshness through the shared `canary-freshness.sh` seam. Closure case M |
| :163 | Partial jq output can drive destructive pruning | FIXED, mechanism refuted | jq does NOT error here: checked live, both `.[][]?` and `index($p)` answer a scalar quietly and exit 0. The consequence is real and worse, so the lock's shape is validated before anything is pruned. Reconcile case I |
| :148 | Removing the final profile assignment prevents pruning | FIXED | The profile universe is the lock union the on-disk profile dirs. Reconcile case J |

`just l` is clean and `just test` is green.

## Effect of merging

Merging makes the tooling exist. It runs nothing: no gate executes, no service is retired, no pull
request is closed, and `chezmoi apply` is never invoked by either script. D1 stays blocked until an
operator runs `cutover-gate.sh 1` deliberately, and the runner stops for two human checkpoints along the
way (approving the retirement manifest, and performing the staged interactive apply, which it refuses to
do itself).

Two things to expect at that first run, both of them the tooling working as intended. Gate 1 will refuse
while the source directory carries stale `.chezmoidata` copies under finished worktrees, because the
rendered view then disagrees with the declaration. It will also refuse while `paseo.json` sits untracked
in the source root, because chezmoi would deploy a file the ledger can never classify. Both refusals
name what to clear.

The one thing merging does unblock is ordering: S12 can now run after this PR, as the plan's
authoritative sequence requires.

## Review guide

Read `scripts/cutover-gate.sh` top to bottom. The shared checks come first (repo handle, base-override
guard, guarded fetch, clean tree, render-versus-source, pin load and validation, two-pin
re-verification, attached-HEAD assertion, `begin_gate`), then gate 1's derivations, then the five gate
bodies in order.

Five places where scrutiny pays off:

- The universe derivation in gate 1 (`derive_label_universe`, `installer_labels_at`) is what catches a
  loaded orphan whose plist was deleted long ago. It adds `--root` to the plan's history queries, since
  without it a label added in the root commit is invisible, and it limits the script-rendered lane to
  deployed paths that actually call `launchctl bootstrap`, which is also where the launchd domain comes
  from. Review found that scan reading this repository's own test fixtures, which is worth re-checking.
- `build_delta_ledger` is where a hunk becomes "landed". It compares tree entries rather than blob ids,
  reads a NUL-delimited path list with rename detection off, and binds every operator classification to
  the exact entry pair it was written for.
- `begin_gate` is the whole marker lifecycle in one place: each gate withdraws its own pass and every
  downstream pass before it does any work.
- The predicate bodies in `verify_against_manifest` match real `launchctl print` output: `state =
  running` plus a pid for persistent jobs, `last exit code = 0` for the conditional-KeepAlive nix-hook,
  and either `run interval =` or an `event triggers` block for scheduled ones. The same function also
  recomputes the managed extras, so the approved snapshot is not the only thing checked.
- Gate 5 names the repository through `--repo=github.com/webdavis/dotfiles`, the documented
  `[HOST/]OWNER/REPO` form, because `gh pr close` has no `--hostname` flag. `GH_REPO` and `GH_HOST` are
  cleared from the child environment rather than merely overridden.

In `scripts/live-reconcile.sh`, the boundary worth checking is what it declines to touch. It prunes
undeclared hermes symlinks into the store but leaves real directories alone, since those are hub-owned
installs, and it reports a missing overlay behind a store symlink instead of writing through it, since
that content belongs to the generation directory or to an app.

Two additions beyond the spec, both testability seams rather than behavior: `CUTOVER_PHASE_A_BASE`
overrides the recorded Phase A base SHA (sandbox repositories cannot contain the real commit, and the
override is refused wherever that commit is reachable), and gate 4's `--window-hours` overrides the
72-hour default, which is my reading of the plan's "days after Gate 1".


webdavis:
Reviewed the gate runner's load-bearing guarantees in code, the 19-item mapping, and the deviations.

Verified directly: the runner never executes `chezmoi apply` (the string appears only in operator-facing instructions; gates checkpoint out at exit 10 and the staged interactive apply is the operator's), and `require_render_matches_source` reads `.chezmoidata/system_packages_autoinstall.yaml` with yq rather than trusting a data-only render, refusing with a diagnostic that cites twpayne/chezmoi#4940 and lists the stale nested copies it found. That refusal firing on this machine today is correct behavior, not a bug: it enforces the open worktree-cleanup blocker at the gate where it matters.

The 19-item table maps each checklist item to a named test, the 17-mutant table has zero survivors across every gate path, and the universe derivation was cross-checked against real repository history including the pre-rename labels. The CI-caught stat portability fix (GNU -c first, BSD -f second, digits or refuse) is pinned by a stub-driven test for each flavor.

Deviations accepted with one flagged for the operator: the committed script invokes `gh` directly rather than the gh-axi wrapper, justified by item 17 specifying gh's resolver semantics and by cutover-day minimalism (gh is the tracked formula; the wrapper needs npx). That is an exception to a standing operator rule embedded in committed code, so it is named here rather than passed silently.

Approved on the Fable pass. Merge waits on the sol review of this same diff.

webdavis:
Final review after the 24-finding fix round.

Every finding was verified against the code before being fixed, none was dismissed, and the one mechanism correction (:163, where jq answers scalars quietly rather than erroring) still fixed the worse underlying consequence. The :279 verification ran the real derivation and caught four phantom system services, two of which were cancelling retirements this procedure exists to perform. The marker lifecycle now withdraws own and downstream passes at gate entry, approval re-validates against a just-derived list at both consuming gates, pins re-verify across the manual apply checkpoint, and retirement requires demonstrated convergence first. Forty-one mutants, zero survivors, with the two first-round survivors leading to stronger tests rather than being argued away.

The standing caveat is correct behavior: gate 1 refuses on this machine today until the stale worktree data copies and the untracked-but-deployable paseo.json are cleared, and each refusal names its remedy.

Approved. Merging.

PR #151, opened 2026-08-04
fix(ssh-hardening): bound, trap and restore what install and reload could not

webdavis:
## Context

`dot_local/bin/executable_ssh-hardening.sh` writes a public-key-only sshd drop-in, verifies the
effective configuration, and puts the tree back the way it found it if that verify fails. PR #142 added
a watchdog to it, because a verify that HANGS is neither a success nor a failure and the transaction
only knew how to answer the second: sshd blocks forever on a named pipe reached through its own Include
glob, so on such a tree the install parked with the drop-in already published and the rollback never
reached.

A review of that merged work found five defects and two places where its new tests pinned nothing.
This branch fixes all seven against current main.

## Summary

- `run_bounded <deadline> <command...>` replaces the verify-only poll, and `--reload`'s syntax check
  (`sshd -t`) and port resolution (`sshd -G`) now run through it. Both were unbounded, and the syntax
  check runs before the verify, so a tree that hangs sshd parked the whole reload before the verify's
  deadline could apply to anything. `dot_local/bin/executable_ssh-hardening.sh`.
- `--rollback`'s recovery proof goes through the same watchdog. It was the fourth unbounded sshd call,
  found while auditing, and it runs after the drop-in has been removed: a hang there leaves the way back
  in open and the operator with no sentence saying so, which is the one moment that sentence matters.
- SIGINT, SIGTERM and SIGHUP during an install now stop the verify's process group, take the same
  rollback a failed verify takes, and re-raise. Measured against main, an interrupt left the new
  drop-in published, the legacy file moved aside, and both verify descendants running.
- The bounded group ignores SIGTTOU and SIGTTIN. Under `stty tostop` a background group is stopped on
  its first write to the terminal, and a stopped process answers `kill -0` exactly like a running one,
  so a healthy install was rolled back and reported as a timeout.
- `cp -Rp` replaces `cp -p` where an existing drop-in is copied aside: a symlinked drop-in is restored
  as a symlink, a dangling one no longer fails the copy outright, and a named pipe no longer blocks it
  forever.
- `SSH_HARDENING_VERIFY_DEADLINE` is capped at 86400 seconds. Past that the tick multiplication
  overflows to a negative budget, the poll runs zero passes, and a healthy verify is stopped at once
  and reported as having waited 2.3e18 seconds.
- `test/e2e/ssh-hardening-verify-watchdog.sh` gains eight cases and two sharper assertions: a floor and
  a 9s ceiling on the wedged run, which is what pins the quarter-second poll interval, and a
  TERM-ignoring wedge, which is what makes the KILL after the stop grace load-bearing.

## How it was verified

Every case is RED without its fix. The e2e suite stops at the first failure, so per-finding evidence
comes from mutations applied to the fixed tree, each reverting exactly one fix. The control (no
mutation) is all 12 cases green, and the whole new suite run against `origin/main`'s script through the
`SSH_HARDENING_SCRIPT` seam fails at case 6, the first case main cannot satisfy.

| Mutation, applied to the fixed tree              | Case that went RED                                                     |
| ------------------------------------------------ | ---------------------------------------------------------------------- |
| M1 syntax check back to an unbounded substitution | 8, through the harness wall clock: "a reload that can hang is itself a regression" |
| M2 install interrupt traps removed                | 6, no INTERRUPTED message and an empty stderr                          |
| M3 `trap '' TTOU TTIN` removed                    | 11, the install under `stty tostop` exited 1                           |
| M4 `cp -Rp` back to `cp -p`                       | 7, the symlink came back as a regular file                             |
| M5 deadline ceiling removed                       | 10, a healthy install rolled back at once                              |
| M6 port resolution back to an unbounded call      | 9, through the harness wall clock                                      |
| M7 `kill -KILL` removed from the group stop       | 5, a pid survived; cases 1 to 4 stayed green                           |
| M8 poll interval 0.25 to 1                        | 1, the wedged run took 15s against the 9s ceiling                      |
| M9 timeout branch worded as a plain parse failure  | 8, which the first version of that assertion did not notice            |
| M10 recovery proof back to an unbounded call       | 12, through the harness wall clock                                     |
| control, unmutated                                | all 12 green, wedged run 5s                                            |

The two coverage gaps needed the opposite demonstration, that the OLD assertion could not see the
defect it was meant to catch. Each of these runs pairs the old assertion with the live defect, and both
are green:

| Old form plus the defect it should have caught       | Result                                      |
| ---------------------------------------------------- | ------------------------------------------- |
| byte-only `file_fingerprint` plus `cp -p`             | green, the symlink was silently lost        |
| old 20s ceiling with no floor plus a 1s poll interval | green, the wedged run took 14s              |

M7 shows the same thing for the KILL: with it deleted, cases 1 to 4 all passed, because the old wedge
was a plain `/bin/sleep` that dies of the first TERM.

Gates: `just l` reports no drift, `just test` exits 0 across all four suites, and CI's lint job passed
on the first commit of this branch.

Two measurements behind the fixes, both on this machine, bash 3.2.57 and macOS 26.2:
`$((2305843009213693952 * 4))` is `-9223372036854775808`; and under a pty with `stty tostop` a child in
its own process group sits in state `T` after writing, while the same child with SIGTTOU ignored sits
in state `S` and its write lands.

### Unbounded call audit

Every external call on the install and reload paths, and what bounds it.

Install:

- `rm -f`, `chmod`, and the three `mv -f` calls: unlink, chmod and rename never open the file, so none
  of them can block on its content.
- `print_config | tee -- <staging>`: opens the staging path for writing, which blocks on a named pipe.
  The same path is removed immediately before. Left as is: recreating it inside that window needs write
  access to the root-owned drop-in directory, which is the trust boundary the whole script assumes.
- the copy of the existing drop-in aside: was unbounded, measured blocking forever on a named pipe.
  Fixed by `cp -Rp`, which recreates a special file instead of reading it.
- the child verify: bounded, and was already.

Reload:

- `sudo -v`: can wait on a password prompt. Deliberate, it is the operator's own prompt, and it runs
  before anything is disturbed.
- the tree observation (`stat`, `cksum`, and the walk's own reads): each open is guarded by a
  regular-file test immediately before it. Residual: a writer swapping a pipe in between the test and
  the open, which again needs write access to the root-owned directory.
- `sshd -t`: bounded here.
- the child verify: bounded.
- `sshd -G` for the probe ports: bounded here.
- `launchctl print`: not a reader of the configuration tree, and no hang has been measured. A wedged
  launchd is a machine-level fault this mode cannot remediate. Left unbounded deliberately.
- `launchctl kickstart -k`: the disruptive step itself. Stopping it midway would leave the service
  somewhere between untouched and stopped, and the readiness prover is what bounds that outcome.
- `ssh-keyscan -T <timeout>`: bounded per attempt by its own flag, and overall by the attempt count.
- `sleep <interval>`: bounded by its argument.

Outside the two paths, found while auditing and fixed here rather than deferred: `--rollback` proves a
password channel is open with two `sshd -G -T -C` calls, and neither was bounded. `check_password_channel`
now owns the temporary file and `resolve_password_channel` asks each question through `run_bounded`, so
the mode reports a stopped proof instead of going quiet. The removal still happens first, and the
failure text now says so, because a re-run is what gets the operator the answer.

## Effect of merging

Nothing on any machine changes when this merges. The script is a chezmoi source file, no `chezmoi
apply` was run, and no test touches `~/.ssh` or the live daemon: every case drives a scratch tree
through the `SSHD_CONFIG_D` and `SSH_HARDENING_SUDO` seams, against stubs.

What changes at the next apply, in operator-visible terms: a `--reload` over a tree that hangs sshd
fails inside the deadline instead of parking, an interrupted install rolls back and says so, an
existing drop-in that is a symlink survives a failed install as a symlink, and a
`SSH_HARDENING_VERIFY_DEADLINE` above 86400 reads as the default 120 rather than as itself.

## Review guide

- The mechanism is `run_bounded`, next to `stop_bounded_group`. Its four call sites are
  `run_verify_child`, step 4 of `reload_sshd`, `resolve_probe_ports`, and `resolve_password_channel`.
  The last two need the command's output, which is why they write to a temporary file: a command
  substitution reads through a pipe and waits for the pipe to close, which is the unbounded wait being
  removed.
- `handle_install_interrupt` and the `INSTALL_ROLLED_BACK` guard belong together. The guard is not
  defensive: a failing step and the interrupt that killed it can both reach the rollback, and a second
  pass over a restored tree reads the absent backup as "nothing was here before" and deletes the file
  the first pass just put back.
- A case label in the e2e file is a stable identifier, not an index into the property list in its
  header and not the running order. Cases 8, 9 and 12 all pin the same property, one bounded call each.
- One thing in the suite is a harness property rather than a script property, and case 6 would prove
  nothing without it: bash gives an asynchronous command SIGINT ignored when job control is off, and a
  signal ignored at entry cannot be trapped, so that case launches under monitor mode. Measured both
  ways before it was written that way.

### DIFF
--- a/dot_local/bin/executable_ssh-hardening.sh
+++ b/dot_local/bin/executable_ssh-hardening.sh
@@ -55,12 +55,14 @@
 #                       reload for as long as it blocks -- the bound limits
 #                       how many times a probe runs, not how long one takes.
 #   SSH_HARDENING_VERIFY_DEADLINE
-#                       seconds the child verify may run before it is stopped
-#                       and reported as a FAILED verify (default 120; see
-#                       VERIFY_DEADLINE_SECONDS for why that bound is safe and
-#                       what an expiry costs). Anything but a positive whole
-#                       number is the default. The override exists for tests,
-#                       which cannot wait out the real bound.
+#                       seconds any ONE sshd call this script waits on may run
+#                       before it is stopped and its failure reported: the child
+#                       verify, and --reload's syntax check and port resolution
+#                       (default 120; see VERIFY_DEADLINE_SECONDS for why that
+#                       bound is safe and what an expiry costs). Anything but a
+#                       positive whole number, or more than one day, is the
+#                       default. The override exists for tests, which cannot
+#                       wait out the real bound.
 #   SSH_HARDENING_ALLOW_MISSING_SSHD
 #                       explicit test seam: when set to a TRUE-ish value AND
 #                       $SSHD_BIN cannot run, --verify skips (exit 0) WITHOUT a
@@ -364,9 +366,28 @@ VERIFY_FAILURES=()
 #
 # A value that is not a positive whole number of seconds is the DEFAULT, never
 # taken at face value: 0 would stop every verify at once, and a non-number would
-# make the arithmetic in run_verify_child a shell error.
+# make the arithmetic in run_bounded a shell error.
+#
+# So is a value past VERIFY_DEADLINE_MAX_SECONDS, and that bound is not
+# decoration. The wait is counted in TICKS, four per second, and bash arithmetic
+# is signed 64-bit: SSH_HARDENING_VERIFY_DEADLINE=2305843009213693952 multiplied
+# by four is exactly -9223372036854775808 (measured on bash 3.2.57), so the tick
+# budget comes out NEGATIVE, the poll loop never runs a single pass, and a
+# perfectly healthy verify is stopped the instant it starts and reported as
+# having timed out after 2.3e18 seconds -- which on the install path rolls a
+# valid install back. The digit-count in the pattern is what keeps the range
+# comparison itself out of overflow territory: a 30-digit value would wrap
+# before `-gt` ever saw it, so it is refused by shape before it is compared.
+#
+# One day is the maximum, and it is already absurd for the job: the measurements
+# above put a real verify at 0.22s and a deliberately oversized one at 5.1s.
+# What the ceiling exists for is arithmetic safety, not policy.
+VERIFY_DEADLINE_MAX_SECONDS=86400
 VERIFY_DEADLINE_SECONDS="${SSH_HARDENING_VERIFY_DEADLINE:-120}"
-[[ $VERIFY_DEADLINE_SECONDS =~ ^[1-9][0-9]*$ ]] || VERIFY_DEADLINE_SECONDS=120
+if [[ ! $VERIFY_DEADLINE_SECONDS =~ ^[1-9][0-9]{0,4}$ ]] ||
+  [[ $VERIFY_DEADLINE_SECONDS -gt $VERIFY_DEADLINE_MAX_SECONDS ]]; then
+  VERIFY_DEADLINE_SECONDS=120
+fi
 VERIFY_STOP_GRACE_SECONDS=2
 # The wait POLLS, because bash has no wait-with-timeout. A quarter second and
 # not a whole one: the check runs before the first sleep, so a whole-second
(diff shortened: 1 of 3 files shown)


webdavis:
Reviewed the fix round against all seven findings plus the audit extensions.

Verified in code: run_bounded is one shared watchdog with the syntax check, port resolution, verify, and the rollback recovery proof all routed through it, and rollback_install is single-shot so a failing step and the interrupt that killed it cannot double-rollback. The disposition table confirms every finding rather than waving any off, and two discoveries exceeded the brief honestly: the rollback itself was destroying symlinked drop-ins (cp -p, fixed with -Rp and a type-carrying fingerprint), and the interrupt scenario was reproduced against main before fixing (published drop-in, live descendants, exit 130).

The evidence standard holds: ten mutations each reverting exactly one fix with a named RED case, blind-form checks proving the old assertions pinned nothing, and the whole new suite RED against main's script through the seam. The left-unbounded list gives a reason per call rather than a blanket claim.

One logged process violation noted for the record: rm on the agent's own mktemp scratch in two throwaway experiment scripts, flagged by the agent itself, no repo or user files touched.

Approved.

PR #152, opened 2026-08-04
fix(osquery): seven review findings against the file-integrity page triage

webdavis:
## Context

The osquery alerter watches this machine and pages when a watched file leaves its root-owned
known-good manifest. PR #146 added file-integrity triage to those pages: three facts (the hash the
manifest records, the hash the file carries now, and whether a recorded Homebrew upgrade plausibly
explains the change) so that a routine vendor update and a tamper stop rendering as the same body.
That work is already merged into `main`.

A review of the merged code by sol (gpt-5.6-sol) came back with seven findings, three rated HIGH.
This branch fixes all seven. Nothing here is new functionality: every change is a defect fix against
code that is on `main` today, and the triage feature keeps the shape it shipped with.

## Summary

- The alerter entry (`dot_local/libexec/osquery/executable_results-alerter.sh`) sourced the optional
  triage helper next to the required libraries, so a helper that had not been deployed yet aborted
  the run under `set -euo pipefail` before any routing, delivery or checkpointing. Every page on the
  machine was blocked until the file came back. That source is now conditional and warns loudly; the
  required libraries stay unguarded and fatal.
- `results-alerter/file-integrity-triage.sh` read the upgrade record with `read -r epoch iso
  <"$record"`, which blocks forever when a readable FIFO stands at that operator-writable path. It
  now refuses anything that is not a readable regular file, and bounds the read at 256 KiB.
- The same helper opened the record twice, once for the run timestamp and once for the package rows.
  The producer publishes by atomic rename, so the two opens could pair one generation's timestamp
  with another generation's transitions. Both shapes now come from one snapshot.
- The same helper decoded rows with `IFS=$'\t' read`, and a tab is an IFS whitespace character, so
  the empty column of an `added` row collapsed and the transition rendered backwards: "14.1.1 ->
  none" for a package that had just appeared.
- `results-alerter/route.sh` validated only that the triage helper returned parseable JSON, so an
  object with wrong member types was attached and then killed the renderer, losing a page the verdict
  had already confirmed. The shape is checked where it is attached now. Separately, that call sent
  the helper's stderr to `/dev/null`, so the launchd log never saw which record could not be read.
- `dot_local/bin/executable_homebrew-weekly-upgrade.sh` wrote the upgrade record only after every
  brew step finished, so a file rewritten early in a run correlated against last week's record. The
  run now publishes its timestamp before the first brew step and fills in the package rows
  afterwards.

## What changed

Six commits, one per logical fix, plus one that corrects the comments describing the record contract
on both sides of it.

| Commit | Finding | File |
| --- | --- | --- |
| `bcf27bb9` | HIGH 1, absent optional helper blocks every page | `executable_results-alerter.sh` |
| `29bff617` | MEDIUM 5, added-row transition reversed | `file-integrity-triage.sh` |
| `a05b08da` | HIGH 2 and MEDIUM 6, blocking record read, two opens | `file-integrity-triage.sh` |
| `10d125eb` | HIGH 3, triage shape not validated | `route.sh` |
| `9875b9f5` | LOW 7, helper diagnostics discarded | `route.sh` |
| `466656a9` | MEDIUM 4, record does not cover its own window | `executable_homebrew-weekly-upgrade.sh` |
| `937a3665` | contract comments, both sides of the record | both of the above |

Three decisions are worth naming, since each closes a finding in a way the review left open.

**The read bound is a byte cap, not a watchdog child.** What blocks forever is a FIFO or a device,
and the regular-file check refuses both at the door; what is left to bound is size, which `head -c`
bounds directly. A watchdog child buys nothing when there is no external process to signal, only this
shell reading a local file. A record over the cap is refused whole rather than parsed from its first
bytes, matching the rule the file already applies to an unrecognised row.

**Fields are decoded by parameter expansion.** `IFS=$'\t' read` cannot express an empty column,
because tab is one of the three IFS whitespace characters and a run of them collapses to a single
delimiter. Splitting on each delimiter in turn keeps the empty column where the producer put it. A
row with too few columns lands a copy of the tail in the state field, which the existing
recognised-state check refuses along with every other shape it cannot describe.

**The homebrew record is published twice per run, both times with the same timestamp.** A `started`
record carrying the run line alone goes out before the first brew step; a `completed` record with the
package rows replaces it after the after-snapshot. The limit is stated in the code: a page fired
mid-run reads a record that dates the run and names nothing, which renders as the existing no-match
line. That line's wording moved from "changed nothing" to "recorded no package change", because the
old wording claims a completed run about one that is still in flight.

## How it was verified

Every fix got a test written first, run RED against unfixed code, then GREEN. Every fix was then
mutation-verified: the fix alone was reverted in an otherwise-fixed tree, the test re-run to confirm
it fails for the right reason, and the fix restored to confirm GREEN.

| Finding | Test | RED without the fix | GREEN with it |
| --- | --- | --- | --- |
| HIGH 1, optional helper | `test/e2e/osquery-alerter-entry.bats` T-ENTRY-optional-triage | entry exits non-zero, no page, cursor never advances | page delivered, cursor advanced, warning on stderr |
| HIGH 2, blocking record | `test/unit/osquery-file-integrity-triage.sh` case 9 | watchdog kills the helper at 10s (exit 137) on a FIFO record | a FIFO and `/dev/zero` both degrade promptly to the could-not-read line |
| HIGH 3, triage shape | `test/unit/osquery-route-pipeline.sh` pass E | `jq: error ... object ({}) cannot be matched`, render exits 5, page lost | page rendered without triage lines, and a well-formed triage object still renders |
| MEDIUM 4, record window | `test/e2e/homebrew-weekly-record.sh` case 4c | no record exists while `brew upgrade` runs | the mid-run record dates the run, at the same epoch as the finalized one |
| MEDIUM 5, added row | `test/unit/osquery-file-integrity-triage.sh` case 8 | `recorded upgrade: ripgrep 14.1.1 -> none` | added, removed and changed each render in the direction the run moved |
| MEDIUM 6, two opens | `test/unit/osquery-file-integrity-triage.sh` case 10 | generation A timestamp paired with generation B transition | the generation the record dates is the one it renders |
| LOW 7, discarded stderr | `test/unit/osquery-route-pipeline.sh` pass F | gate stderr empty | the helper's path and reason reach the alerter log |

Two of those tests need a note on how they are driven, because neither condition arrives by waiting.

The blocking-record case runs the helper under a watchdog, since a plain call would wedge the suite
rather than fail it. The bound is a killer subshell, not a poll loop, so the passing path sleeps for
nothing: `wait` returns the instant the helper exits and the killer is torn down unused.

The two-opens case drives the interleaving deterministically rather than racing it. The helper reads
its clock between the two opens it used to do, so a `date` shim on PATH is a hook that fires exactly
there, and it publishes a second generation over the record. The shim leaves a marker and the test
asserts that marker, so the case cannot pass by quietly failing to drive the swap.

`main` advanced during the work (a cask declaration and a brew-cache test race fix, neither touching
these files), so it is merged into the branch and the gates were re-run on the merged tree. `just l`
reports no drift. `just test` exits 0 across all four suites with zero `not ok` lines.

## Effect of merging

Nothing changes on this machine when this merges. Every source file here is chezmoi source state, and
none of it reaches `$HOME` until an apply, which is held for the D1 cutover. The live alerter keeps
running the code deployed today, including the blocking read and the unguarded source, until then.

After that cutover the behavior differences are: a page fires even when the triage helper is missing;
a record that is not a readable regular file costs a page its correlation line instead of costing the
page; a triage object the renderer cannot print is dropped rather than taking the page with it; the
helper's degradation reasons appear in the alerter's launchd log; an `added` package renders its
transition in the right direction; and a file changed during the weekly Homebrew run correlates
against that run rather than against the previous week's.

## Review guide

Start with `dot_local/libexec/osquery/results-alerter/file-integrity-triage.sh`, which carries four
of the seven findings. The region to read is the top of `_file_integrity_upgrade_line`: the type
check, the bounded snapshot, and the row decoding are three separate fixes, and each states in a
comment what it refuses and why.

Then `dot_local/libexec/osquery/results-alerter/route.sh` at the triage attachment point, where the
shape check and the stderr change are adjacent but independent.

The producer change in `dot_local/bin/executable_homebrew-weekly-upgrade.sh` reads best against its
consumer: `write_upgrade_record` takes a phase, the two call sites bracket the brew work, and
`UPGRADE_RECORD_EPOCH` is read once near the top so both records date the same moment.

For the tests, `test/unit/osquery-file-integrity-triage.sh` cases 8 through 10 are the new ones, and
case 10 is the one whose mechanism is worth understanding before trusting it.

### DIFF
--- a/dot_local/bin/executable_homebrew-weekly-upgrade.sh
+++ b/dot_local/bin/executable_homebrew-weekly-upgrade.sh
@@ -81,6 +81,16 @@ if [[ -n $UNATTENDED_LOG_AVAILABLE ]]; then
   LOG_ENTRY_HEADER="$(unattended_log_entry_header "$LOG_SUCCESS_MARKER")"
 fi
 
+# The upgrade record's timestamp, read ONCE here and used by both the record this
+# run publishes before it starts and the one it finalizes after (see
+# write_upgrade_record). It dates the moment the run BEGAN, which is what makes
+# the record cover the window it describes: the epoch is what the consumer does
+# arithmetic on and the ISO string is what it renders, and two `date` calls could
+# disagree about which run they belong to.
+UPGRADE_RECORD_EPOCH=""
+UPGRADE_RECORD_ISO=""
+read -r UPGRADE_RECORD_EPOCH UPGRADE_RECORD_ISO < <(date -u '+%s %Y-%m-%dT%H:%M:%SZ' 2>/dev/null) || true
+
 # The relay script by ABSOLUTE path. The LaunchAgent's PATH is
 # /opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin, with no
 # ~/.local/bin in it, so a bare `relay.sh` would never be found under launchd and
@@ -165,8 +175,26 @@ __homebrew_package_snapshot() {
   esac
 }
 
-# write_upgrade_record -- persist what this run moved, so something days later
-# can ask whether an upgrade plausibly explains a file that changed.
+# write_upgrade_record <started|completed> -- persist what this run moved, so
+# something days later can ask whether an upgrade plausibly explains a file that
+# changed.
+#
+# PUBLISHED TWICE, ONE RUN. `started` writes the run line alone before the first
+# brew step; `completed` rewrites it with the package rows once the after
+# snapshot is taken. Both carry the SAME timestamp, so the two are one record at
+# two levels of detail rather than two runs. It was written only at the end
+# before, and that left the whole upgrade window uncovered: a watched file
+# rewritten in the first seconds of a run pages while the newest record on disk
+# is still the previous week's, so the correlation answers "no recorded upgrade
+# in the last 3 days" about a file an upgrade is moving right then.
+#
+# THE LIMIT, since this is a lead and its limits are what keep it honest: a page
+# fired mid-run reads a record that dates the run and names NOTHING, because
+# nothing has been compared yet. That renders as the no-match line, which states
+# what the run recorded rather than claiming a mapping; it is the same sentence a
+# genuinely empty week produces. Naming the packages mid-run would mean diffing
+# the Cellar on every page, which is the fork-brew-from-the-alert-path this
+# deliberately does not do.
 #
 # WHO READS IT. The osquery file-integrity page fires when a watched file leaves
 # its known-good manifest, and a vendor update and a tamper used to render the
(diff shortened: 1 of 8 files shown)


webdavis:
Reviewed the seven-finding round.

Verified in code: the optional triage helper is now a guarded readable-test source with a loud warning naming the missing path, so the router's declare -F fallback is finally reachable, while the required libraries stay fatal. The disposition table reproduces every finding against main before fixing it, and the mutation table carries the actual RED output per finding rather than an assertion that one exists.

Three judgment calls accepted with their reasons: the bounded read as a regular-file check plus a byte cap rather than a watchdog child (nothing external to signal; the blocking classes are refused at the door); the two-generation race driven by a deterministic PATH shim with a marker the test asserts (the interleaving cannot be produced by waiting, and the shim cannot silently not fire); rm in test teardown matching the suite convention since trash is not in the flake shell. The IFS whitespace-collapse diagnosis for the added-row reversal explains both the bug and why prior coverage missed it, and the added coverage now spans all three row kinds.

The extra contract-comment commit and the reworded no-match line are correct in-branch scope under the terminal ruling, and the two information-only notes are information, not deferred work.

Approved.

PR #153, opened 2026-08-04
fix(update-skills): close the review findings against the merged idle-gate work

webdavis:
## Context

`~/.local/bin/update-skills.sh` is the weekly job that keeps the cross-harness skills store
(`~/.agents/skills`) matching the roster this repo declares in `dot_agents/custom-skill-lock.json`. It
builds a candidate generation, publishes it with one atomic exchange, and fans the store out to Claude
Code and hermes as symlinks. PR #143 deleted its harness-activity idle gate (which had deferred every
scheduled slot on a machine in daily use) and delisted the `website-to-video` skill.

A review of that merged work found seven defects, one ranked HIGH, plus three comments still describing
mechanisms that no longer exist. This PR fixes them on top of `main`. Two other HIGH findings about
generation pinning are deliberately not addressed here: they belong to the Rust work and the operator
owns that trade.

## Summary

- A lock holding more than one JSON document no longer passes the roster gate in
  `dot_local/bin/executable_update-skills.sh`. `jq -e '<filter>'` reads a file as a stream and answers
  for the LAST document, so a roster with `{}` appended passed every check on that empty object while the
  readers still used the first document's tables.
- The same stream-acceptance hole is closed in four sibling reads, found by auditing every `jq` read of a
  lock or state file in the script: the fan-out's own `claudeDelivery` check, the candidate npx lock
  validation, the two npx-lock reconcile inputs, and the failure-streak state file.
- A `claudeDelivery` key containing a control character is refused. The reader is line-oriented, so the
  single key `"gh-axi\nhumanizer"` was two names to it, and convergence reaped both skills' Claude links.
- A `--dry-run` that refused part of its work now exits non-zero, and lock contention on the last
  scheduled slot of the week now alerts instead of ending the week silently with no update and no stamp.
- The updater's own hash is captured once at start-up, so a `chezmoi apply` landing mid-run can no longer
  make an old parent stamp the new updater's identity (which made every later slot early-exit on work
  that never ran).
- The run now names live content the roster no longer declares, a delisted skill's store directory and
  the `skillOverrides` entry left with it, and removes neither.

## What changed

One commit per finding, plus the map refresh. A separate commit carries the treefmt follow-ups (shfmt
realigning a comment column and one reasoned shellcheck directive).

| Commit | Finding |
| --- | --- |
| `refuse a multi-document lock` | HIGH: the schema gate answered from the last document; four sibling reads audited and fixed with it |
| `refuse a claudeDelivery key holding a control character` | a newline in a key forged a second exemption and reaped two links |
| `exit non-zero when a dry run refused part of its work` | a refused preview looked like a clean one |
| `alert when the last scheduled slot defers over contention` | a contended last slot ended the week silently |
| `capture the updater's identity once, at start-up` | an old parent stamped the new updater's hash |
| `name the live content the roster no longer declares` | delist leftovers were invisible |
| `make the stale-stamp case prove a rebuild happened` | the test passed on any silent early exit |
| `correct three comments about gates that no longer exist` | the plist slot comment, the weekly-record test, the install test |
| `refresh the graphify map` | the committed map still indexed deleted functions |

Two decisions worth naming, both recorded in comments next to the code:

**Contention on the last slot stays a deferral, and alerts.** Nothing was attempted, so it is not a
required-phase failure and the exit stays the retryable 75. What the last slot adds is the alert, on the
route exhaustion already uses. Waiting for the lock instead was rejected: a holder can run for many
minutes, and blocking the last slot on it trades a visible miss for an invisible stall inside a launchd
job nobody watches.

**The delist report removes nothing, by ruling.** It compares the store against the lock's `tiers` table
(which covers exactly the roster, so vendored and app-owned entries stay quiet) and names what is left
over. The `skillOverrides` half is reported only for a name the store also carries, which is the delist
shape: Claude Code takes overrides for skills this vertical does not deliver, and naming those would be a
weekly false positive.

## How it was verified

Every fix has a test that fails without it. RED was measured by stashing the source change and running
the test against the pre-fix script, GREEN by the same test on the fixed one. Each fix was then mutated
against an unmutated control, and the killing test recorded.

| # | Fix | Mutation applied | Test that went RED |
| --- | --- | --- | --- |
| 1 | roster schema gate slurps | judge the last document (`length >= 1 and (.[-1] \| ...)`) | `test/integration/update-skills-roster-failclosed.sh` case 4b |
| 2 | fan-out `claudeDelivery` check slurps | same shape on that filter | `test/integration/update-skills-converge.sh` (the dry run offered to restore the link) |
| 3 | candidate npx lock validation slurps | back to `jq -e .` | `test/unit/update-skills-json-stream-reads.sh` |
| 4 | reconcile inputs slurp | back to `jq -e .` | same (the reconcile failed the candidate instead of falling back) |
| 5 | streak file read slurps | back to `jq -e .` | same (the state file was left holding two documents) |
| 6 | control-char keys refused (fan-out) | `test("[[:cntrl:]]")` never matches | `update-skills-converge.sh` (the dry run previewed reaping both links) |
| 7 | control-char keys refused (gate) | the same, in the gate | `update-skills-converge.sh` (the run completed instead of failing closed) |
| 8 | dry run exits non-zero on refusal | `exit 1` back to `exit 0` | `update-skills-converge.sh` |
| 9 | last-slot contention alerts | exhaustion check forced false | `test/e2e/update-skills-lock-contention.sh` (no exhaustion logged) |
| 10 | ...and only on the last slot | exhaustion check forced true | the same file (a mid-week slot alerted) |
| 11 | updater hash captured at start | read the path at write time again | `test/integration/update-skills-stamp.sh` (the stamp recorded an updater the run never executed) |
| 12 | unrostered content reported | call site removed | `test/integration/update-skills-delist.sh` |
| 13 | ...including the stale override | override lookup forced false | the same file (it named the directory, not the override) |
| 14 | ...and only for unrostered names | membership test inverted | the same file |
| 15 | stale-stamp case proves a rebuild | one silent early exit added to the updater | the new assertion failed; **the old assertion passed**, which is the finding |

The multi-document scenario is pinned end to end as reported: the fixture roster with `claudeDelivery`
emptied and `{}` appended made the pre-fix full run print
`converge: created .../.claude/skills/nonclaude`, recreating a deliberately de-delivered link. After the
fix the run refuses, the link stays absent, and no stamp is written.

### The jq-stream audit

Every `jq` read of a lock or state file in the script, and where it stands now:

**Fixed (5).** `__gen_roster_schema_ok` (the roster lock, the HIGH);
`__update_skills_claude_delivery_malformed` (the roster lock at the point of use, the only gate
`--dry-run` reaches); `__gen_validate_candidate` (the candidate `.skill-lock.json`, whose later readers
ask `.skills | has($n)` and would have answered from the last document);
`__gen_reconcile_candidate_npx_lock` (both inputs reach `jq --argjson`, which refuses a stream outright,
so a stream turned a fallback into a hard failure); `__gen_update_failure_streaks` (a stream
re-incremented every slot and grew the file by a document per run, and a bare scalar was fatal under
`errexit`).

**Audited, unchanged, with the reason.** Every other roster read in a mutating mode goes through the
run-private snapshot taken after the gate, so the gate covers them; in `--dry-run` they read the live
lock, and the fan-out check now refuses a multi-document lock before any of them can mislead, with a
non-zero exit. `check_fork_drift` already required a single document
(`__gen_fork_lock_single_document`), which is where the shape came from. `generation.json` and the
exchange-in-flight marker are written by this script with `jq -n`, and a stream makes their id
comparisons unequal, which fails toward "rebuild" or "drop the marker", never toward accepting.
`.clawhub/origin.json` is publisher-controlled, but its value is already scrubbed of control characters
before it is printed, so a multi-document marker collapses to one line and cannot forge a row in the
weekly record.

### Gates

`just l` clean, `just test` green across all four suites, with `main` merged in at `8fe01fed`.

## Effect of merging

Nothing changes on this machine. Every file here is chezmoi source state and no apply runs before the D1
cutover, so the deployed `~/.local/bin/update-skills.sh` and the loaded LaunchAgent keep their current
behaviour until then.

Three things happen at the first apply after cutover. The weekly run gains the refusals and the two new
reports. The updater's bytes change, so the current week un-stamps and the next scheduled slot rebuilds
once. And the plist's comment edit changes the rendered file, so `run_onchange_after_63` (which keys on
that file's hash) reloads the LaunchAgent, bootout then bootstrap, with the same schedule it has now.
The delist report will name `website-to-video`'s leftovers on that run; it removes neither, and removing
them stays a hand operation.

### On the two out-of-scope HIGHs

Neither is touched, and nothing here makes either more likely. The exchange still resolves generations
per lookup rather than pinning one per invocation, and the serialize lock still does not cover the
out-of-band HyperFrames writer. One second-order note for completeness: capturing the updater hash means
a mid-run replacement now leaves a real mismatch, so a later slot rebuilds where it used to early-exit,
which is one more exchange in exactly the case where the old behaviour was recording work that never ran.

## Review guide

Read `dot_local/bin/executable_update-skills.sh` first; the fixes are small and each carries its reason
above it. Three are worth the most attention:

- `__gen_roster_schema_ok` and `__update_skills_claude_delivery_malformed`, the two slurped gates. The
  question to ask is whether any reader can still reach a lock that neither of them has judged.
- the contention branch at the lock acquisition, for the semantics: a deferral, exit 75, and an alert
  only on the last scheduled slot.
- `report_unrostered_live_content`, for the false-positive argument: it fires on a store name absent from
  `tiers`, and mentions `skillOverrides` only for such a name.

For the tests, `test/integration/update-skills-converge.sh` carries the multi-document and forged-key
scenarios with both the full run and the preview, and `test/unit/update-skills-json-stream-reads.sh` pins
the three sibling reads at the function level in about a second.

### DIFF
--- a/Library/LaunchAgents/com.webdavis.update-skills.plist.tmpl
+++ b/Library/LaunchAgents/com.webdavis.update-skills.plist.tmpl
@@ -22,11 +22,15 @@
   <key>RunAtLoad</key>
   <false/>
   <!-- 24 hourly Monday retry slots (00:00..23:00), generated programmatically.
-       A slot that finds a harness with recent activity defers and leaves no
-       weekly success stamp, so the next slot retries; once a slot succeeds, the
-       rest no-op on the stamp. The 23:00 slot is the LAST (a deferral there
-       alerts loudly). Keep the last hour in sync with UPDATE_SKILLS_LAST_SLOT_HOUR
-       in update-skills.sh. -->
+       A slot writes the weekly success stamp only when it fully succeeds, so a
+       required-phase failure, or a deferral because another run holds the
+       serialize lock, leaves the week unstamped and the next slot retries; once
+       a slot succeeds, the rest no-op on the stamp. (There is no harness-activity
+       gate any more: a run under a live session is safe, and the gate deferred
+       every slot on a machine in daily use.) The 23:00 slot is the LAST, where
+       both a failure and lock contention alert loudly, having nothing behind
+       them. Keep the last hour in sync with UPDATE_SKILLS_LAST_SLOT_HOUR in
+       update-skills.sh. -->
   <key>StartCalendarInterval</key>
   <array>
 {{- range $hour := until 24 }}
--- a/dot_local/bin/executable_update-skills.sh
+++ b/dot_local/bin/executable_update-skills.sh
@@ -876,7 +876,8 @@ __gen_roster_schema_ok() {
       and ((.clawhubTracked // {}) | to_entries
         | all((.value | type == "object")
           and nonempty_string(.value.slug) and nonempty_string(.value.registry)))
-      and ((.claudeDelivery // {}) | to_entries | all(.value == "none")))
+      and ((.claudeDelivery // {}) | to_entries | all(.value == "none"))
+      and ((.claudeDelivery // {}) | keys | all(test("[[:cntrl:]]") | not)))
   ' "$1" >/dev/null 2>&1
 }
 
@@ -2370,6 +2371,13 @@ converge_dir() {
 # other direction is unthinkable here: a jq that failed would otherwise mark
 # every roster skill undelivered and step 2 of converge_dir would reap the
 # ENTIRE Claude fan-out on one bad read.
+#
+# ONE NAME PER LINE is this reader's whole contract, which is why the two
+# validators refuse a key holding a control character. A key spelled
+# "gh-axi\nhumanizer" is one exemption to the schema and TWO names here, and the
+# second one costs a link: converge_dir removes every updater-owned link outside
+# the desired set, so one forged name reaps a delivery nobody de-delivered. No
+# store name contains a control character, so nothing legitimate is refused.
 __update_skills_claude_undelivered() {
   [[ -f $CUSTOM_SKILL_LOCK ]] || return 0
   jq -r '.claudeDelivery // {} | to_entries[] | select(.value == "none") | .key' \
@@ -2396,7 +2404,8 @@ __update_skills_claude_delivery_malformed() {
   jq -e -s 'length == 1 and (.[0] |
       (has("claudeDelivery") | not)
       or ((.claudeDelivery | type == "object")
-        and (.claudeDelivery | to_entries | all(.value == "none"))))' \
+        and (.claudeDelivery | to_entries | all(.value == "none"))
+        and (.claudeDelivery | keys | all(test("[[:cntrl:]]") | not))))' \
     "$CUSTOM_SKILL_LOCK" >/dev/null 2>&1 && return 1
   return 0
 }
--- a/dot_local/bin/executable_update-skills.sh
+++ b/dot_local/bin/executable_update-skills.sh
@@ -848,21 +848,35 @@ GEN_ROSTER_HASH=""          # sha256 of the snapshot at run start
 # a present-but-non-object table and any value other than the string "none"
 # (absent means the default, delivered), so a malformed delivery table refuses the
 # run loudly instead of silently restoring a de-delivered skill.
+#
+# The gate reads the lock SLURPED, and the -s is the load-bearing part.
+# `jq -e '<filter>' file` reads a STREAM of values and evaluates the filter once
+# per document, so its exit status is the LAST document's verdict while every
+# extractor downstream still reads them ALL. A roster with a second top-level
+# `{}` appended therefore passed every check here on the trailing empty object
+# (an absent table is legal) and the run went on to read the real tables out of
+# the first document: with claudeDelivery emptied in that first document, the
+# undelivered-name reader returned nothing and convergence RECREATED a
+# deliberately absent Claude link before stamping the week a success. Slurping
+# and requiring exactly one value is the single-value test; the same shape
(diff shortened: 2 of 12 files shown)


webdavis:
Reviewed the fix round and the audit.

Verified in code: the roster schema gate now slurps and requires exactly one document before any member check runs, the same shape the settings quarantine established, and the HIGH was reproduced live before the fix (the pre-fix run created the exempted Claude link from a two-document lock). The audit's fixed-and-left split is the right kind: five reads fixed, each remaining read justified by who writes it or what refuses a stream downstream, including the --argjson case where a stream already hard-fails.

Judgment calls accepted with their reasons: last-slot contention alerts rather than retries (a lock holder can run minutes; blocking trades a visible miss for an invisible stall), and the unrostered-content report writes to the run log without a relay push, because a weekly buzz for state the operator may deliberately keep would train the alert away. Both semantics are commented where they live. The out-of-scope generation-pinning HIGHs stay untouched for the Rust port, and the one second-order effect of capturing the updater hash at start-up (a mid-run replacement now causes a real rebuild instead of an early exit) is stated rather than discovered later.

Fifteen mutations killed with named tests, both alert directions pinned, stale-stamp blindness measured before being fixed.

Approved.

PR #154, opened 2026-08-04
fix(osquery): read every monitor state file as one JSON document, not a stream

webdavis:
## Context

This repository is a chezmoi dotfiles tree, and part of what it deploys is an osquery-based security
monitoring pipeline: small bash scripts run by launchd that watch the machine (firewall, Gatekeeper,
screen lock, Tailscale Funnel exposure, the health of the pipeline itself) and page when something
changes for the worse. Each of those monitors keeps a small JSON state file between runs so it can tell
a fresh problem from one it has already reported.

An audit of how those scripts read JSON found one bug class repeated across four of them. jq accepts a
stream of concatenated documents, and these reads never said that a state file must be exactly one of
them. `jq -e .` takes its exit status from the LAST document, so a file holding two valid documents
passed the corruption check. `.key // default` then runs once per document, so a query over such a file
emits one line per document, and a document that lacks the key contributes an empty line that command
substitution strips. All four scripts already documented that a corrupt state starts fresh; this makes
that promise true rather than incidental.

The honest blast radius: these state files live in `~/.local/state` at mode 0600, every monitor rewrites
its file at the end of a healthy tick, and none of this grants an attacker capability they did not have
(anyone who can write the state file can write it with one document just as easily). What a doubled file
bought was silence, for as long as it lasted: a live Funnel that read as steady state, a manifest tamper
whose page was suppressed by a resurrected marker, and a poller that wrote the doubling back into its own
baseline.

## Summary

- `dot_local/libexec/osquery/executable_uptime-watchdog.sh`: the state file is read once through a
  slurped "exactly one object" guard instead of `cat` plus `jq -e .`; the seven later reads already query
  that in-memory value, so nothing re-reads the file.
- `dot_local/libexec/osquery/executable_firewall-gatekeeper-monitor.sh`: the posture baseline gets the
  same whole-file rule `load_controls` already applies to the controls file, which also disarms the
  fold-in on the refused-controls path that used to persist a multi-document baseline.
- `dot_local/libexec/osquery/executable_tailscale-monitor.sh`: the funnel baseline is slurped to one
  object before its value is read, so a concatenated file can no longer read as a steady exposure.
- `dot_local/libexec/osquery/executable_allowlist.sh`: the per-line label read now uses `fromjson?`, the
  same one-value-per-line rule `results-alerter/allowlist-verdict.sh` applies when it consults the file,
  and a source line that is not exactly one JSON tuple refuses the whole mutation instead of being edited
  around.
- Three comments that described guards the code did not have now describe the mechanism that exists,
  including the `jq empty` parse check in the Tailscale monitor, which stays deliberately permissive and
  now says why.
- Nine bats cases across four suites: a concatenated-file case per script (two for the allowlist writer,
  which has two curation verbs), a single-document round-trip case for each of the three monitors, and a
  comment/blank-line round-trip case for the allowlist writer.

## What changed

The shape is the same at all four sites, copied from the one place in the repo that already had it
(`load_controls` in the posture monitor):

```
jq -ce -s 'if (length == 1 and (.[0] | type == "object")) then .[0] else error("not one object") end'
```

Read the file once, validate the whole file, then query the validated value. Per site:

- uptime-watchdog: The corruption gate accepted two documents. Downstream, the numeric guards
  (`// -1`, `// 0`, each behind a regex) reject the two-line values a stream produces, but the two
  `// ""` fingerprint reads do not: a trailing `{}` collapses the fan-out back to one clean 64-hex value.
  The one that matters is `paged_fingerprint`, the marker that says this exact divergence was already
  reported. Adopted from a corrupt file, it silences the manifest-audit page for that tamper permanently,
  and the manifest audit is the only detector for a hard-linked, symlinked, or chmodded script (osquery
  watches paths, so those generate no file event).

- firewall-gatekeeper-monitor: `prev_fw`, `prev_gk` and `prev_sl` are read with `// empty`, which
  emits nothing for a document lacking the field, so a baseline followed by a second document collapsed
  to a clean trio and passed `prev_valid`. The per-control priors read the same way. That armed the
  amplifier: with the controls file refused, `jq -c '. + $trio' <<<"$prev_json"` runs over the
  multi-document prior, emits one object per document, and `write_state` persists the whole stream, so
  the doubling outlives the tick that read it and the next tick re-pages every declared control as a
  first observation.

- tailscale-monitor: `.funnel // empty` read straight from the state file collapsed the same way, so
  a live Funnel matched an "active" baseline and the public-exposure page never fired. The status-output
  parse check keeps `jq empty`: that input is one capture of one CLI run rather than a file, and the
  classifier below re-checks the shape and gaps on anything it does not recognize. Its comment says that
  now instead of reading like a shape check.

- allowlist: The writer read each line as a stream, so a line holding two concatenated tuples
  reported two labels, matched neither, and survived the rewrite, while `-d` printed "denied" and
  refreshed the manifest and `-a` appended a second tuple beside it. The consumer already reads one value
  per line with `fromjson?` and ignores such a line, so nothing was suppressed by it either way; what
  needed fixing is a writer that edits around a corrupt source and reports success. It now reads labels
  by the consumer's rule and refuses the mutation outright: non-zero, reason and path on stderr, nothing
  deployed, no manifest refresh. The committed seed file passes the rule (7 entry lines, none refused).

## How it was verified

The audited mechanics were re-verified against the installed jq (1.8.2) before any edit: `jq -e .`
returns 0 for two concatenated objects and for `false` followed by an object; `// -1` over two documents
yields two lines; `// ""` and `// empty` over `{obj}{}` yield one; the slurp guard exits 5 on both
concatenated files and 0 on the single one; `fromjson?` yields nothing for a two-document line.

Every case was written first and observed RED for the audited reason, then GREEN. Each guard was then
reverted on its own and its suite re-run beside the unmutated single-document control.

| Guard reverted | Corruption case | Control case | Observed failure |
| --- | --- | --- | --- |
| watchdog slurp to `cat` + `jq -e .` | RED | GREEN | the state persisted a `paged_fingerprint` lifted from the corrupt file, and the confirmed tamper never paged |
| poller slurp to `cat` | RED | GREEN | the baseline persisted as two documents, carrying `filevault` from the distrusted prior |
| tailscale slurp to `.funnel // empty` | RED | GREEN | 0 pages with a live public Funnel |
| allowlist refusal to `return 0` | RED (both verbs) | GREEN | `-d` exited 0 reporting "denied" and `-a` exited 0 reporting "allowed", over a corrupt source |
| allowlist reader to the stream read | GREEN | GREEN | survives: the refusal short-circuits before the label read ever sees a multi-document line (see the review guide) |

Suite state: `just l` reports 0 files changed, and `just test` passes all four suites.

## Effect of merging

Nothing on the live machine changes on merge. These are chezmoi-managed source files under
`dot_local/libexec/osquery/`, so the deployed copies in `~/.local/libexec/osquery/` are replaced only by
a `chezmoi apply`, which is held until the D1 cutover. No test ran against live osquery state: every case
runs in its own temp HOME with stubbed osqueryi, launchctl, tailscale, and dispatch.

Once applied, monitor behavior is unchanged for every state file that is a single JSON object, which is
every file these monitors write themselves. A file that is not one starts fresh, or for the allowlist
writer refuses the mutation, rather than being partly believed.

## Review guide

- The four guards are one line each, and the same line. Compare against `load_controls` at
  `executable_firewall-gatekeeper-monitor.sh:463`, which is where the shape came from.
- For the uptime-watchdog, check the change against its seven reads: they all query `$prev_state` and
  none re-reads `$STATE`.
  `grep -n 'prev_state' dot_local/libexec/osquery/executable_uptime-watchdog.sh` shows the whole set.
- The allowlist refusal is the one behavior change beyond validation, and it is deliberate: reading
  labels the consumer's way is not observable on its own, because a two-document line matches no label
  either way and survives the rewrite either way. Without the refusal, `-d` keeps reporting a removal it
  did not make. That is also why the reader mutation survives in the table above: both predicates enforce
  the same one-value-per-line rule, and the refusal runs first. It cannot fire on the shipped file either:
  `test/integration/osquery-page-allowlist-seed.sh` already pins that every entry line there is one NDJSON
  tuple with exactly the four fields.
- Each corruption case is paired with a pretty-printed (indented, multi-line, still one document)
  round-trip case, because the rule counts documents and a plausible wrong implementation counts lines.

### DIFF
--- a/dot_local/libexec/osquery/executable_allowlist.sh
+++ b/dot_local/libexec/osquery/executable_allowlist.sh
@@ -67,13 +67,31 @@ take_allowlist_write_lock() {
   "$lockf_bin" -s 9
 }
 
-# The JSON label of an allowlist line, or empty for a comment/blank/non-JSON line.
+# The JSON label of an allowlist line, or empty for a comment/blank/non-tuple line.
+# `-R` reads the LINE as a raw string and `fromjson?` parses it, so a line counts as
+# an entry only when it holds EXACTLY ONE JSON value. That is the rule
+# allowlist_verdict applies when it consults the deployed file, and matching it is
+# what keeps writer and consumer agreeing on what an entry is: read as a STREAM
+# instead, a line holding two concatenated tuples reports two labels, matches
+# neither, and survives a rewrite the consumer honors it for nothing.
 # 9>&- so this jq (spawned under the write lock) never inherits the lock fd - see the
 # fd-inheritance note on take_allowlist_write_lock.
-entry_label() { jq -r '.label // empty' 9>&- <<<"$1" 2>/dev/null || true; }
+entry_label() { jq -Rr 'fromjson? | .label // empty' 9>&- <<<"$1" 2>/dev/null || true; }
+
+# line_is_one_tuple <line> -- status 0 when the line holds exactly one JSON object.
+# The refusal predicate below; same one-value-per-line rule as entry_label, so the
+# two can never disagree about which lines are entries.
+line_is_one_tuple() { jq -Re 'fromjson? | type == "object"' 9>&- <<<"$1" >/dev/null 2>&1; }
 
 # Rewrite an allowlist file, preserving comment/blank lines and dropping any tuple for
 # <label>. Reads <file>, writes the filtered result to stdout (a no-op if it is absent).
+#
+# A non-comment line that is not exactly one JSON tuple REFUSES the whole rewrite
+# (nonzero, reason on stderr). Editing around it silently is the failure this
+# closes: the label read cannot match such a line, so the line survives while the
+# command reports a removal it did not make, or an add lands a second tuple for a
+# label the file already mentions. A corrupt source is fixed by hand, not curated
+# around.
 _without_label() {
   local drop="$1" file="$2" line
   [[ -f $file ]] || return 0
@@ -84,6 +102,11 @@ _without_label() {
         continue
         ;;
     esac
+    if ! line_is_one_tuple "$line"; then
+      printf 'refused: the allowlist source holds a line that is not a single JSON tuple; repair %s by hand before curating it\n' \
+        "$file" >&2
+      return 1
+    fi
     [[ "$(entry_label "$line")" == "$drop" ]] && continue
     printf '%s\n' "$line"
   done <"$file"
@@ -244,7 +267,10 @@ allow_label() {
   fi
   local temp
   temp=$(mktemp 9>&-)
-  _without_label "$label" "$source_file" >"$temp"
+  if ! _without_label "$label" "$source_file" >"$temp"; then
+    rm -f "$temp" 9>&-
+    exit 1
+  fi
   jq -cn --arg label "$label" --arg path "$rel_path" --arg program "$rel_prog" --arg sha256 "$sha" \
     '{label:$label, path:$path, program:$program, sha256:$sha256}' 9>&- >>"$temp"
   publish_allowlist "$temp" "$source_file" || exit 1
@@ -273,7 +299,10 @@ deny_label() {
   fi
   local temp
   temp=$(mktemp 9>&-)
-  _without_label "$label" "$source_file" >"$temp"
+  if ! _without_label "$label" "$source_file" >"$temp"; then
+    rm -f "$temp" 9>&-
+    exit 1
+  fi
   publish_allowlist "$temp" "$source_file" || exit 1
   printf 'denied: %s\n' "$label"
 }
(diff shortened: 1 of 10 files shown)


webdavis:
Reviewed the seventeen-site round.

The strongest part of this PR is that it corrected its own brief: the audit claimed the watchdog's numeric reads were the hazard, and the implementation demonstrated they were already inert behind their regex guards, while the real live bug was the two fingerprint reads collapsing on a document lacking the key, which permanently suppressed the manifest-audit page once a fingerprint had been paged for. The test drives that exact suppression. The firewall amplifier was observed persisting a two-document baseline before the fix rather than asserted, and the tailscale zero-pages-on-live-funnel scenario reproduced.

Judgment calls accepted: the allowlist writer refusal beyond the prescribed reader change, since the reader change alone is unobservable (proven by the one unkilled mutant, which is reported in the open with its reason: the refusal short-circuits before the label read, and both predicates enforce one rule); the nine-case count; and the steady-deviant reseed after the first control survived the line-counting mutant, which is the dead-test check doing its job. The three corrected comments now describe mechanisms the code actually has.

CI is green and mergeable_state is clean. Approved and merging.

PR #155, opened 2026-08-04
fix(plugin-record): harden the weekly plugin record, and sweep the jq stream reads

webdavis:
## Context

`~/.local/bin/report-plugin-updates.sh` is the weekly record of what Claude Code's own plugin
auto-update changed on this machine. Claude Code refreshes marketplaces and their installed plugins at
startup by itself and leaves no trace, so this helper reads
`~/.claude/plugins/installed_plugins.json`, compares it against last week's reading, and posts one
entry to the `#unattended-upgrades` channel. It landed in PR #141 along with the plugin declarations
in the settings modify-template. A review of that work raised ten findings against it.

Seven are in the reporter, two are in `update-skills.sh` and its roster test, and one is a repo-wide
sweep of a bug class that has now turned up in three consecutive slices: `jq` reads a file as a
SEQUENCE of JSON documents, so a doubled or appended document is accepted everywhere a single one is
meant. The sweep covered every `jq` invocation under `dot_local/bin/`, `dot_local/libexec/`,
`.chezmoiscripts/` and `scripts/`, and its full disposition is below.

Two groups of sweep sites belong to other branches: PR #153 fixed `update-skills.sh` while this branch
was in flight (main is merged in, so this branch carries that fix), and the osquery helpers are held by
`fix/osquery-jq-stream-guards`, which has not merged yet. The disposition table says which sites came
from where.

## Summary

- The reporter no longer accepts an inventory it cannot trust: a file holding two JSON documents is
  refused, and an install record whose shape it cannot read refuses the run instead of dropping out of
  the reading and being announced as a plugin removal
  (`dot_local/bin/executable_report-plugin-updates.sh`).
- An inventory with no user-scope records is now reported as the removals it describes, rather than
  raising `plugin-state-unreadable` every week from then on.
- Snapshot handling is fixed in three ways: a snapshot path that is not a regular file is refused
  loudly, the snapshot is renamed into place instead of copied over, and a reading that cannot be
  persisted alerts on the priority route instead of only reaching the local log.
- The baseline is seeded at deploy time by
  `.chezmoiscripts/run_onchange_after_69-load-report-plugin-updates-launchagent.sh.tmpl` through a new
  `--seed-baseline` mode, so the first transition after an apply is reported rather than absorbed into
  the baseline.
- Four fixes outside the reporter: `update-skills.sh` names the de-delivered link its apply-time mode
  leaves behind (no removal is scripted), `test/unit/skills-roster-fanout.sh` ties its delivery
  tripwire to the runtime read rather than to a string the schema validator also carries,
  `dot_local/bin/executable_assert-hermes-superpowers-routing.sh` refuses a lock holding more than one
  document, and `dot_local/bin/executable_find-and-remove-json-objects.sh` passes its value to jq as
  data, which also fixes a compile error that made it fail on every file it matched.

### The reporter, `dot_local/bin/executable_report-plugin-updates.sh`

The inventory reader is now slurped (`jq -rs`), so the document count is checked before anything is
read out of it. That is a semantic fix and not only a status one: an `error()` raised while processing
one document of a stream does not make jq exit non-zero when another document succeeds, so the
degraded-file refusals underneath could not refuse anything on a doubled file. `-e` came off in the
same pass, because it exits 4 when a filter produces no output, and no user-scope records is an answer
rather than a failure. Every install record is now shape-checked before the scope filter rather than
after it.

`write_snapshot()` stages the reading as a sibling in the state directory and renames it over the
snapshot, so a write interrupted halfway cannot leave a short file that the next run reads as a batch
of new plugins. Both callers go through it. A snapshot path that exists and is not a regular file is
refused before the comparison, since a directory there reads as "no baseline yet" on every run
forever. Both persistence failures now alert.

The script runs under `set -euo pipefail`. The conversion is only safe once the failures it means to
tolerate are guarded where they happen, so the relay call, the host lookup and the library's
bookkeeping write say so explicitly. That last one was not tolerated on purpose: the library warns and
returns 0 whatever happened, by design, so a run that delivered its entry and could not record itself
exited 0 and left the next entry measuring its gap from a run that never happened.

`--seed-baseline` records the first reading and stops. The loader calls it after deploying the helper
and before bootstrapping the agent. The ordering requirement is stated in the loader: the same apply
enables the marketplace auto-updates this record reports on, so the seed has to happen in that apply,
and nothing between the settings write and the seed may start Claude Code. Seeding is idempotent, so a
routine apply cannot re-baseline over a change nobody has reported yet.

### Outside the reporter

`converge_dir()` in `update-skills.sh` warns, with the absolute path, about an updater-owned link that
is no longer delivered while running `--install-only`. That mode is what runs at apply time and it
removes nothing, so the link survives until a full weekly run reaps it and the harness sees two
sources under one name for that whole window. No removal is scripted here.

`test/unit/skills-roster-fanout.sh` scopes its grep to `__update_skills_claude_undelivered`'s own
body. The expression it looks for also appears in the roster schema gate, which only validates the
table, so a file-wide grep stayed green while the runtime filter was repointed at a misspelled table.

`assert-hermes-superpowers-routing.sh` checks the lock is one JSON object before any of its three
reads. On a doubled lock the pair walk collected the union of both documents and the slash-command
skill became whichever document carried it, and both drive an in-place rewrite of the mirror.

`find-and-remove-json-objects.sh` passes its value with `--arg`. It was spliced into the jq program in
single quotes, which jq has no such thing as, so the filter was a compile error and the helper failed
on the first file it matched, every time. The selector stays program text because a jq path is what
the tool takes, so its shape is bounded to a plain dotted path and refused before any backup directory
is created.

## How it was verified

Every fix was written test-first: the assertion was run against the unmutated tree and observed to
fail for the stated reason, then the fix was applied and the assertion observed to pass. Each fix was
then mutated back individually against an otherwise-unmutated control, and the control re-run green
afterwards.

| # | Finding | Test | RED evidence before the fix | Mutation applied after | Assertion that fired |
|---|---------|------|------------------------------|------------------------|----------------------|
| 3 | Doubled inventory accepted | `test/e2e/report-plugin-updates-record.sh` section 9 | exit 0, and the entry read `2 of 6 tracked entries changed (exa 4.0.0 -> 5.0.0, exa 4.0.0 -> 9.9.9)` | slurp removed | a doubled inventory exited 0 |
| 4 | Records filtered before validation | section 10 | exit 0, entry read `1 of 3 tracked entries changed (exa (removed))` from a `scop` typo | shape checks removed | mistyped scope key exited 0 |
| 2 | Empty user scope read as unreadable | section 11 | exit 1 with an empty `jq said:` reason | `-e` restored | an inventory with no user-scope records exited 1 |
| 5 | Directory at the snapshot path | section 12 | exit 0, "recording this reading as the baseline", copy written inside the directory | guard removed | a snapshot path that is a directory exited 0 |
| 6 | Snapshot overwritten in place | section 13 | inode unchanged across a delivering run | `write_snapshot` reverted to `cp` | the snapshot was written in place rather than renamed over |
| 7 | Persistence failure only logged | section 14 | exit 1 with no alert on the relay | alert removed | a run that could not persist its baseline alerted nobody |
| Q | `set -uo pipefail` | section 15 | a run that could not record itself exited 0 | marker check removed | a run that could not record its own success exited 0 |
| 1 | Baseline seeded at first scheduled run | section 16 and `test/unit/report-plugin-updates-plist.sh` | `--seed-baseline` exited 2 as an unknown argument | seed call removed from the loader | the loader never seeds the baseline |
| Q | Hour unpinned in the plist test | `test/unit/report-plugin-updates-plist.sh` | retiming to 14:00 stayed green | plist Hour set to 14 | the agent no longer fires at 13:00 (Hour 14) |
| 10 | Tripwire matched the validator | `test/unit/skills-roster-fanout.sh` | the file-wide grep stayed green with the runtime filter repointed at `.claudeDelivry` | runtime filter repointed at `.claudeDelivry` | `__update_skills_claude_undelivered()` no longer reads `.claudeDelivery // {}` |
| 9 | Leftover link left unnamed | `test/integration/update-skills-converge.sh` | the apply-time run said nothing about the link it left | warning removed | the apply-time run left a de-delivered link behind without naming it |
| S | Routing lock read as a stream | `test/integration/hermes-superpowers-routing.sh` section 10 | refused for the wrong reason, reporting a malformed slash-command skill that was two copies of a valid one | guard removed | the refusal does not say the lock is not one JSON value |
| S | jq value spliced as program text | `test/unit/find-and-remove-json-objects.sh` | `jq: error: syntax error, unexpected INVALID_CHARACTER`, helper exited 1 without touching a file | `--arg` reverted to interpolation | the helper exited non-zero on a plain value |
| S | Selector accepted unbounded | same file | new guard, no prior behaviour | selector pattern widened to `.*` | a refused selector still created `backup_20260804115633` |

The one behaviour with no runnable check is rename atomicity itself, which cannot be interrupted on
demand. The inode assertion stands in for it: `cp` keeps the inode, `rename` replaces it, so the test
pins the mechanism that provides the guarantee rather than the guarantee.

### The jq sweep

Every `jq` invocation under `dot_local/bin/`, `dot_local/libexec/`, `.chezmoiscripts/` and `scripts/`
was read and classified: 160 invocations across 32 files. `scripts/` contains none.

Verified behaviour that the sweep rests on, measured in this repo with jq 1.8.2:

- `jq -e 'F' file` on a two-document stream takes its exit status from the LAST document alone, so a
  file whose first document is malformed passes a gate when the trailing one is valid.
- `jq -r 'F' file` on a two-document stream emits rows from BOTH documents.
- An `error()` refusal inside a stream does not set a non-zero exit when another document succeeds, so
  an error-based guard cannot refuse multi-document input at all without slurping first.

Fixed here:

| Site | Consequence of a stream | Fix |
|------|--------------------------|-----|
| `report-plugin-updates.sh`, the inventory read in `plugin_snapshot()` | duplicated fingerprint rows, a fabricated weekly transition, and the degraded-file refusals bypassed | slurp, require exactly one document |
| `assert-hermes-superpowers-routing.sh`, the routing-pair walk | contradictory pairs merged into the rewritten map | one slurp-and-count guard before all three reads |
| `assert-hermes-superpowers-routing.sh`, the `slashCommandsSkill` read | picks whichever document carried the key, choosing which `SKILL.md` gets rewritten | same guard |
| `assert-hermes-superpowers-routing.sh`, the pairs handed to perl | prints a null error and still exits 0 with the other document's pairs, then rewrites every file in the mirror | same guard |

Audited and not changed here, with reasons:

| Site or group | Verdict |
|---------------|---------|
| `update-skills.sh`: roster schema gate, fan-out `claudeDelivery` check, the two npx reconcile inputs, candidate lock validation, failure-streak state | Needs the fix, and got it in PR #153 (`fix/update-skills-sol-findings`, commit dfd8c986) while this branch was in flight. That work is merged into main and into this branch, so the fix is present here without being duplicated. |
| `update-skills.sh`: 21 roster reads (795, 1339, 1353, 1373, 1437, 1499, 1504, 1656, 1685, 1696, 1833, 1865, 2063, 2070, 2357, 2441, 2500, 2526, 2578, 2596, 2629) | Guarded. In every mutating mode these read the validated run-private roster snapshot, which the gate above admits. |
| `update-skills.sh`: 7 fork-table reads (3089, 3103, 3114, 3144, 3324, 3348, 3349, 3350) | Guarded by the pre-existing `__fork_lock_single_document` (3074), which already slurps and counts. |
| `update-skills.sh`: self-written state (580 `generation.json`, 980, 981, 1027 exchange marker) | A truncating writer plus a rename produces a truncated file on a crash, never a doubled one, and each of these already tolerates an unreadable read. |
| `update-skills.sh`: the published npx lock and `origin.json` reads in `__update_skills_change_snapshot` | Written by third-party CLIs, and both feed the advisory weekly change record, so the worst outcome is a wrong line in a record rather than a mutation. The lock is validated before publish by PR #153's candidate check; `origin.json` is not, and is reported here as information rather than changed, since PR #153 audited this file. |
| `update-skills.sh`: 565, 739, 1531, 1912 | `jq -n` constructs JSON and reads no input. |
| `dot_local/libexec/osquery/**`, 17 sites | Needs the fix, covered by the sibling branch `fix/osquery-jq-stream-guards`. Not duplicated here. |
| `dot_local/libexec/osquery/**`, results-log readers | Legitimate stream: `osqueryd.results.log` is newline-delimited JSON by design. |
| `.chezmoiscripts/run_before_12-quarantine-unparseable-claude-settings.sh` | Already correct, and is the reference implementation this PR follows. |
| `dot_local/bin/executable_relay-codex-hooks.sh:43` | Guarded at line 24, whose `jq -es` read is what admits the value. The comment above line 43 claimed the check itself tested singleness; the comment is corrected, the code is not. |
| `dot_local/bin/executable_relay-agent.sh` JSONL read | Reads raw with `-R -s`, so document framing does not apply. |
| Hook-payload stdin reads (`claude-user-prompt-start`, `claude-stop-pulse`, `relay-codex-hooks`), single-response CLI pipelines (`smart-lights`, `hue-pulse`, `herdr-jump`, `claude-audit`), the `.chezmoiscripts` captures | One document per invocation by the producer's contract. |

One defect of a different class surfaced during the sweep and is fixed here because it sits in the
same file set: `find-and-remove-json-objects.sh` interpolated both of its arguments into the jq program
text. See the table above.

### Gates

`just l` reports no drift (386 files, 0 changed). `just test` passes.

## Effect of merging

Nothing is applied by merging. Every change here is source state, and this repo has not applied since
before the cutover, so the live machine is untouched until the next `chezmoi apply`.

At that apply, three things happen that did not before. The loader records a plugin baseline before
bootstrapping the agent, so the first weekly entry reports what moved since the apply instead of
starting from whatever the machine looked like the following Monday. The apply-time `--install-only`
pass prints a warning naming any updater-owned link that is no longer delivered, including
`~/.claude/skills/last30days`, which the operator then removes by hand; nothing removes it
automatically, and the full weekly run reaps it as before. And the reporter refuses inputs it used to
accept, so a degraded `installed_plugins.json` produces an alert and no entry rather than an entry
nobody should trust.

`find-and-remove-json-objects.sh` starts working. It has been a compile error on every invocation that
matched a file, so nothing depended on its previous behaviour.

## Review guide

Read the reporter as one file: `dot_local/bin/executable_report-plugin-updates.sh` carries seven of the
findings, and the reasoning for each sits above the code it explains.

Two judgement calls worth a second opinion:

- `--seed-baseline` is called from `run_onchange_after_69`, which fires when the plist or its own
  content changes. That covers the first deploy, which is the window the finding is about, and it does
  not re-seed if the state directory is wiped later. The alternative, a `run_before_` script seeding
  strictly before the settings write, would close a sub-second window that requires Claude Code to
  start mid-apply. The loader was chosen because it already exists.
- Finding 8 and the `update-skills.sh` half of the sweep are not in this PR's own commits. PR #153
  fixed both delivery gates and three more reads while this branch was in flight, so duplicating them
  here would have meant two conflicting edits to the same functions. Main is merged in, so those fixes
  are present in this branch; the only change this PR makes to `update-skills.sh` is in
  `converge_dir`.

Test placement follows the repo's suites: reporter behaviour is end-to-end because it runs the real
helper against fixture state, the schedule and argument checks are unit, and the two convergence cases
are integration.

### DIFF
--- a/.chezmoiscripts/run_onchange_after_69-load-report-plugin-updates-launchagent.sh.tmpl
+++ b/.chezmoiscripts/run_onchange_after_69-load-report-plugin-updates-launchagent.sh.tmpl
@@ -7,6 +7,34 @@ set -euo pipefail
 
 mkdir -p "$HOME/.local/log/plugins"
 
+# SEED THE BASELINE HERE, and mind the order. The record compares this week's
+# plugin versions against a stored previous reading, and whichever run records
+# that first reading reports nothing itself. If the FIRST scheduled run is the
+# one that does it, everything Claude Code's own auto-update changed between this
+# apply and that Monday is baked into the baseline and reported never: deploy on
+# Tuesday, a plugin moves on Wednesday, Monday records the new version as the
+# starting point and says nothing moved.
+#
+# ORDERING REQUIREMENT: this apply is also the apply that enables marketplace
+# auto-updates (the extraKnownMarketplaces entries in
+# private_dot_claude/modify_settings.json, applied with the other targets before
+# any run_after script), so the seed has to happen in the SAME apply, and nothing
+# between that settings write and this line may start Claude Code. It must also
+# stay AFTER the helper itself is deployed and BEFORE the agent is bootstrapped
+# below, so the first scheduled run always finds a baseline waiting.
+#
+# Seeding is idempotent: an existing baseline is left alone, so re-running this
+# on every plist change costs nothing and cannot swallow a change.
+HELPER="$HOME/.local/bin/report-plugin-updates.sh"
+if [[ -x $HELPER ]]; then
+  # Never fatal. A chezmoiscript that exits non-zero aborts the whole apply, and
+  # a machine with no plugin inventory yet has no first transition to lose.
+  "$HELPER" --seed-baseline ||
+    printf 'report-plugin-updates: the baseline could not be seeded at apply time; the first scheduled run will record it instead, and anything that moves before then goes unreported\n' >&2
+else
+  printf 'report-plugin-updates: %s is not deployed yet, so no baseline was seeded\n' "$HELPER" >&2
+fi
+
 PLIST="$HOME/Library/LaunchAgents/com.webdavis.report-plugin-updates.plist"
 TARGET="gui/$(id -u)/com.webdavis.report-plugin-updates"
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -509,6 +509,13 @@ names no other delivery mechanism and reads no other lock, per the operator's st
 `"none"` is the only legal value, and a malformed table refuses the run rather than failing open, in
 every mode including `--dry-run`.
 
+**Retiring an EXISTING link is manual, and the run says so.** Deleting the chezmoi declaration does not
+remove a `~/.claude/skills` link already on the machine (chezmoi never deletes a target it no longer
+manages), and the apply-time `--install-only` pass is additive, so it removes nothing either. The link
+therefore survives until the next FULL weekly run reaps it, and for that window Claude Code sees two
+sources under one name. `converge_dir` now WARNS with the absolute path in the additive mode, naming what
+it is leaving behind for the operator to delete. No removal is scripted, by operator ruling.
+
 **Tier model (the lock's `tiers` table):** every roster skill is `core` (8) or `on-demand` (27). Core
 skills auto-load in every harness; on-demand skills stay installed everywhere but load only when
 explicitly invoked: in Claude Code via `skillOverrides.<name> = "user-invocable-only"`, one
@@ -765,10 +772,25 @@ shared entry shape and the reasoning behind it).
 - **Schedule:** `com.webdavis.report-plugin-updates`, Monday 13:00, `RunAtLoad=false`, logging to
   `~/.local/log/plugins/report-updates.log`. It passes `--scheduled`, and only a scheduled run posts,
   moves the snapshot or advances the marker; a manual run prints the comparison and changes nothing.
-- **First run** records a baseline and posts nothing. **A quiet week still posts**, naming zero changes,
-  because a clean week and a dead LaunchAgent otherwise produce identical silence. **An unreadable state
-  file posts no record at all** and alerts on the priority route instead, since the only change list it
-  could build from a file it cannot read is a false "nothing changed".
+- **The baseline is seeded at APPLY time**, by the loader chezmoiscript calling `--seed-baseline`, not by
+  the first scheduled run. The apply that deploys this record is the apply that turns the marketplace
+  auto-updates on, so a baseline first recorded the following Monday absorbs everything Claude Code
+  changed in between and reports it never. Seeding is idempotent, an existing baseline is left alone, so
+  a routine apply cannot re-baseline over a change nobody has reported yet. It is also best effort and
+  never pages: a machine with no readable inventory yet seeds nothing, says so, and leaves the baseline
+  to the first scheduled run.
+- **A first run with no baseline** records one and posts nothing. **A quiet week still posts**, naming
+  zero changes, because a clean week and a dead LaunchAgent otherwise produce identical silence. **An
+  inventory it cannot read posts no record at all** and alerts on the priority route instead, since the
+  only change list it could build from a file it cannot read is a false "nothing changed". That set
+  includes a file holding more than one top-level JSON document (jq accepts a stream, and both copies
+  would reach the reading) and an install record whose shape it cannot interpret (dropping one out of the
+  reading announces the plugin as REMOVED). An inventory with no USER-scope records is NOT in that set:
+  it is a real reading of a real machine, and its removals are reported.
+- **The snapshot is replaced by rename**, never written in place, so a run interrupted halfway cannot
+  leave a short file that the next run reads as a batch of new plugins. A snapshot path that exists and
+  is not a regular file refuses the run, and a reading that cannot be persisted alerts, because both
+  otherwise produce a machine that reports nothing while looking healthy.
 
 ### Happy Daemon (Remote Agent Control)
 
--- a/dot_local/bin/executable_assert-hermes-superpowers-routing.sh
+++ b/dot_local/bin/executable_assert-hermes-superpowers-routing.sh
@@ -75,6 +75,20 @@ if [[ ! -f $LOCK ]]; then
   exit 0
 fi
 
+# EXACTLY ONE JSON VALUE, checked once, before any read below trusts the file.
+# jq reads a file as a STREAM and runs its filter per document, so a doubled or
+# appended lock parses without complaint and each read here answers differently:
+# the pair walk collects the union of both documents, and the slash-command skill
+# becomes whichever document happened to carry it. Every one of those feeds an
+# in-place rewrite of the files in the mirror, so the routing this script exists
+# to assert would come from a document nobody chose. Slurping and counting is the
+# single-value test, the same shape run_before_12-quarantine-unparseable-claude-settings.sh
+# uses on ~/.claude/settings.json.
+if ! jq -e -s 'length == 1 and (.[0] | type == "object")' "$LOCK" >/dev/null 2>&1; then
+  log "refusing: $LOCK is not a single JSON object; a multi-document or unparseable lock would route the mirror from a document nobody chose"
+  exit 2
+fi
+
 # Read the routing pairs. Every name feeds a regex and a written line, so
 # validate the shape hard: lowercase kebab, legacy names superpowers-prefixed.
 legacy_names=()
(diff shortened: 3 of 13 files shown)


webdavis:
Reviewed the final fix round and the sweep.

Verified in code: the loader seeds the baseline after the helper deploys and before the agent bootstraps, with the ordering requirement written where it lives, so the first post-cutover plugin transition is reported instead of silently absorbed. The routing lock now refuses a multi-document or unparseable file before any read, one guard covering all three consumers, and the refusal message says what a stream would have cost. The slurp uses == 1 rather than <= 1 with the right reason: a zero-document inventory is degraded for a reporter, not acceptable.

Finding 8 handled correctly: recognized as already fixed by the sibling that merged mid-flight, verified present via the main merge, not duplicated. The sweep's two extras are in-scope catches, the never-worked find-and-remove tool above all: a filter that was a compile error on every invocation is exactly the artifact-reports-success class, and it is now fixed with --arg plus a bounded selector. The information-only note about origin.json stays information, owned by the file's own audit.

Approved.

PR #156, opened 2026-08-04
docs: share one global ruleset across harnesses and rebuild CLAUDE.md

webdavis:
## Context

This is a [chezmoi](https://www.chezmoi.io/) dotfiles repository. Two instruction files steer agents
working in it: a global one that loads into every session on this machine, and the repo `CLAUDE.md` that
loads whenever an agent works here. Both had drifted. The global ruleset reached Codex only through a
hand-edited, untracked `~/.codex/AGENTS.md`, which still named haiku as the commit-message model and `gh`
as the git TUI months after the Claude copy was corrected. The repo file had grown to 922 lines and
carried claims nobody had rechecked against the code in weeks.

This is the last slice before the cutover, so the fix is one pass over both files: give the two harnesses
one shared source, move the conditional detail into runbooks, and re-verify every remaining claim by
reading the code it describes.

## Summary

- `.chezmoitemplates/global-agent-rules.md` is a new shared partial that `~/.claude/CLAUDE.md` and
  `~/.codex/AGENTS.md` both pull in with `includeTemplate`, between a pair of `shared-rules` markers.
  `private_dot_codex/AGENTS.md.tmpl` puts the Codex global file under chezmoi for the first time.
- `test/integration/global-instruction-parity.sh` renders both global targets with the real chezmoi and
  byte-compares what lands between the markers, so inlining the block in one target, editing one copy, or
  dropping a marker fails the build.
- Five runbooks under `docs/runbooks/` now hold the material that only matters while working on one
  subsystem: the skills store, the Claude Code settings template, the git hooks, the local daemons, and
  the macOS defaults runners.
- `CLAUDE.md` drops from 922 lines to 490: identity, commands, an architecture map, the conventions, and
  a table pointing at the runbooks.
- Every factual claim in both files was re-verified against current `main`. The correction that changes
  what an agent does: `chezmoi apply --exclude=templates` was documented as vault-free and is not, since
  it does not skip `modify_` templates and two of those call `keepassxc`.

## How it was verified

- `just L` (the `nix flake check` treefmt drift gate): `all checks passed!`
- `just test` (all four suites): exit 0.
- `test/integration/global-instruction-parity.sh` mutation-verified against an unmutated control. Adding
  one line inside the codex markers fails with a diff of the two shared blocks; replacing the include
  with inlined text fails the mechanism guard; deleting both markers from the Claude target fails the
  empty-block guard.
- `test/unit/skills-roster-fanout.sh` rule 10 now reads `CLAUDE.md` and the skills runbook as one corpus.
  Mutation-verified four ways: a wrong npx count in the runbook, a wrong on-demand tier size, two files
  disagreeing on the roster size, and a dropped sentence that would have retired its own check.
- Four parallel agents fact-checked the old `CLAUDE.md` section by section against the code, one ledger
  line per claim with a file:line citation. The stale ones are itemized in the commit messages.
- Both rendered global targets inspected end to end to confirm the shared block and each harness tail
  land where intended.
- `grep` for em-dashes and en-dashes across every changed file and every commit message: none. The
  detector was control-tested against a file that has both.

## Effect of merging

Merging changes no live state. Nothing here runs at merge time, and the repo currently applies nothing
to the machine.

At the next `chezmoi apply`, three targets change. `~/.claude/CLAUDE.md` is rebuilt from the shared
partial plus its Claude tail. `~/.codex/AGENTS.md` becomes chezmoi-managed and its hand edits are
replaced, including the stale haiku and `gh` lines; the codegraph-owned block is carried forward in the
tracked file, reflowed to 105 columns with its em-dashes removed, so a later codegraph rewrite shows up
as a diff to fold back rather than a silent overwrite. `~/.codex` itself gains a chezmoi-managed 0700
directory entry, which matches its current mode.

Nothing else in `$HOME` is touched: the runbooks and `CLAUDE.md` are source-only files that
`.chezmoiignore` keeps out of the target state.

## Review guide

Read `.chezmoitemplates/global-agent-rules.md` and the two thin wrappers first; that is the whole
mechanism, and the rest of the branch is prose. Then `test/integration/global-instruction-parity.sh`,
where the marker extraction and the anchor guard are the parts worth arguing with. The anchor is what
stops two marker-less targets from comparing equal and passing.

`CLAUDE.md` reads better as a new file than as a diff. The section most worth a second pair of eyes is
Chezmoi operations, which now contradicts guidance this repo has carried for months: check the two
`modify_` templates named there with `grep -n keepassxc` before trusting it.

The five runbooks are the largest part of the diff and the least load-bearing, since they mostly carry
text that already existed. The corrections folded in on the way are itemized in the runbook commit
message, which is the efficient way to spot-check them.

`test/unit/skills-roster-fanout.sh` has one behavioral change: rule 10 reads two documents instead of
one, and a disagreement between them is now its own failure mode.

### DIFF
--- a/.chezmoitemplates/global-agent-rules.md
+++ b/.chezmoitemplates/global-agent-rules.md
@@ -71,9 +71,6 @@ Require per-invocation confirmation. Blanket "yes" doesn't carry over.
   alternatives (`fd`, `rg`, `sd`). Modern tools are for interactive use; scripts need boring reliability.
 - ISO 8601 timestamps. On macOS use `gdate -Is` or `date -u +"%Y-%m-%dT%H:%M:%SZ"`, BSD `date` lacks
   `-Is`.
-- `[[ ]]` for tests, never `[ ]`. Lists are arrays, never space-separated strings. Never
-  `for x in $(command)`; read with `while IFS= read -r x; do ...; done < <(command)`. Build JSON with
-  `jq -n --arg`/`--argjson`, never string interpolation.
 
 ## Git commits
 
@@ -86,14 +83,10 @@ Use the `conventional-commits` skill. A user-wide `prepare-commit-msg` hook prep
 
 ## Pull request descriptions
 
-Bodies use five sections in this order: `## Context`, `## Summary`, `## How it was verified`,
-`## Effect of merging`, `## Review guide`. Posting is gated on a self-review against the anti-AI-pattern
-checklist (no em-dashes, no negative parallelism, no inflated significance, no vague attribution, no
-rule-of-three padding, no third-person talk about the audience).
-
-`~/.claude/commands/pr.md` is the single source for both the section contract and the checklist. Read it
-before drafting a body and follow it there; never restate or fork those rules elsewhere. Keep a posted
-body current when later review rounds change the facts.
+`~/.claude/commands/pr.md` is the single source for the body's section contract and for the
+anti-AI-pattern self-review that gates posting. Read it before drafting a body and follow it there; never
+restate or fork those rules elsewhere. Keep a posted body current when later review rounds change the
+facts.
 
 ## Task tracking
 
--- a/.chezmoitemplates/global-agent-rules.md
+++ b/.chezmoitemplates/global-agent-rules.md
@@ -1,6 +1,7 @@
-<!-- Evergreen file. General principles, not transient state. -->
-
-# Global Rules
+<!-- Shared global ruleset, included verbatim by ~/.claude/CLAUDE.md and ~/.codex/AGENTS.md. Edit here,
+     never in a harness copy: test/integration/global-instruction-parity.sh byte-compares the rendered
+     block across both targets and fails when they diverge. Harness-specific rules go in the including
+     file, below the shared block. -->
 
 ## Collaboration style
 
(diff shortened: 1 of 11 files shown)


webdavis:
Reviewed the final slice.

Verified in code: both wrappers pull the identical includeTemplate between identical markers, and the parity test byte-compares the extracted blocks with an anchor check so two marker-less targets cannot vacuously compare equal. The factual corrections are the valuable core, above all the one this program itself leaned on: --exclude=templates does not skip modify-templates, so the "safe, no KeePassXC prompt" claim is corrected to match the measured behavior, alongside the missing tenth vault target and the wrong source-dir description. Roster rule 10 reading both files as one corpus is the right repair for a pin whose subject moved.

Deviations accepted: the global refactor and shared partial as one commit (splitting leaves a tree that does not build), the reflowed codegraph block with its fold-back note, the softened just-a claim (verifying it would have meant an apply, which nothing may run), and the partial architecture diagram, the lane graph covers the mechanism and the full per-skill lifecycle graph remains unbuilt, stated openly rather than pretended.

One workflow consequence to know, not a defect: the global CLAUDE.md is now a template, so the automation apply path that excludes templates will no longer refresh it; it updates via the interactive apply, which is the same path its KeePassXC-adjacent siblings already take.

Approved.

PR #158, opened 2026-08-05
fix: port six unshipped remnants the cutover ledger surfaced

webdavis:
## Context

This repository is the chezmoi source for one machine. Its `main` branch is a reimplementation: the work
was rebuilt slice by slice off an older `integration` branch, which is now frozen and still describes
what is actually deployed. The cutover gate that compares the two trees keeps a ledger of expected
differences, and six things in it turned out not to be expected at all. They exist on `integration`, they
are live on the machine, and no slice ever landed them on `main`. Nobody decided to drop them; they fell
into the seams between slices.

This is the last pull request before the operator's D1 cutover apply, which is why they are due now:
after D1 `main` is what writes the machine, so anything only `integration` carries is gone.

## Summary

- `dot_bashrc.tmpl`: the long-running command notifier calls `~/.local/bin/relay.sh` instead of raising
  its own `alerter` banner, so a finished command gets presence routing (at the desk, banner only; away,
  add the phone push), the Discord paper trail, and a banner whose click focuses the herdr pane.
- `test/integration/bashrc-long-command-relay.sh`: a new test pinning that a long shell command reaches
  relay with its state and its pane, local-only below five minutes and fanned out above. This is what the
  notifier lacked. Every other relay producer has a test and kept its wiring through the rewrite.
- `dot_hermes/private_dot_env.tmpl`: the hermes gateway's environment file, with five API credentials and
  three Discord IDs rendered from KeePassXC, so rotating one in the vault reaches hermes and a rebuilt
  machine gets credentials at all.
- `.chezmoidata/hermes.yaml` plus `.chezmoiscripts/run_onchange_before_25-install-hermes-agent.sh.tmpl`:
  the pinned hermes-agent ref and its only consumer, which is what a fresh machine needs to have hermes.
- `.chezmoiscripts/run_after_59-hermes-config-migrate.sh.tmpl` and
  `.chezmoiscripts/run_once_after_60-moshi-hook-setup.sh.tmpl`: two apply-time scripts `main` referred to
  but did not carry. `run_after_72` already said it runs after the 60-series moshi-hook setup.
- Two small ones: `dot_config/worktrunk/config.toml` puts worktrees back under
  `~/.herdr/worktrees/<repo>/<branch>`, which is herdr's layout and where the live checkouts are, and
  `private_dot_claude/modify_settings.json` takes `effortLevel = xhigh` under chezmoi control.

## How it was verified

- The test is the mutation proof for the notifier. It was written first and ran RED against `main` for
  the right reason (relay called zero times, not a harness error), GREEN after the change. Eleven
  separate mutations of the notifier each turn it RED again, with an unmutated control GREEN before and
  after the run: reverting to the `alerter` call, dropping `--local-only` from the first tier, adding it
  to the second, dropping the pane, dropping the project, deleting the exit-code state derivation,
  lowering the first threshold to 20s, removing the agent CLIs from the skip list, leaving a bare
  `--pane` at the end of the argv, leaving an `alerter` banner beside relay, and handing the light pulse
  a hardcoded success.
- One mutation is expected to survive, and does: relay-agent.sh omits `--pane` entirely when herdr has
  not set it, where the notifier passes an empty value. Both keep the flag and its value paired, which is
  the contract, so the test accepts either. What it refuses is the dangling flag: `--pane` with nothing
  behind it, which relay's parser warns about and drops.
- `effortLevel` is mutation-verified the same way: deleting its `setValueAtPath` line fails
  `test/unit/claude-enabled-plugins.sh` on 48 assertions across both target platforms.
- No `chezmoi apply` was run in any form, and keepassxc was never invoked. The templates were verified by
  rendering instead: `just s` renders every discovered `.chezmoiscripts/*.sh.tmpl` and shellchecks the
  result, `test/integration/rendered-template-coverage.sh` puts the two new scripts on the right side of
  the classifier (the migrate script renderable and covered, the moshi-hook one excluded because it reads
  the vault), and `test/unit/secret-source-files-declare-private-mode.sh` confirms chezmoi maps
  `dot_hermes/private_dot_env.tmpl` to `~/.hermes/.env` at mode 0600.
- `just l` clean, `just test` green across all four suites.

## Effect of merging

Merging changes nothing on the machine. All of this is chezmoi source state and none of it reaches
`$HOME` until the operator's D1 apply. Two of the new files read the vault (the hermes environment
template and the moshi-hook setup script), so they need an interactive apply with KeePassXC unlocked and
`chezmoi apply --exclude=templates` skips them.

After that apply, the visible change is the shell notifier: it stops raising its own banner and goes
through relay, so a command finishing while nobody is at the desk reaches the phone at the five-minute
tier and lands in Discord, and the banner can be clicked to focus the pane the command ran in.
`~/.hermes/.env` is rewritten from the vault, so every entry it names has to exist and be populated
first, and worktrunk starts creating worktrees under `~/.herdr/worktrees/`; existing worktrees are not
moved.

## Review guide

- Start at `test/integration/bashrc-long-command-relay.sh`. The notifier detaches every call it makes and
  a detached grandchild cannot be waited for, so instead of a settle time the driver runs with fd 9 on
  the command substitution's pipe and the stubs record onto it. The substitution returns at EOF, which is
  when the last of them has exited, so the cases expecting nothing are as deterministic as the ones
  expecting a record. That is the part worth reading closely.
- Where `main`'s version was kept over `integration`'s: the first tier fires at 30 seconds, not
  integration's 60, because 30 is the threshold CLAUDE.md documents; the skip list keeps `hermes`, which
  integration's copy predates; `MESSAGING_CWD` renders from `.chezmoi.homeDir` instead of a literal home
  path, matching how the LaunchAgent templates build absolute paths; and the environment file's header
  says the gateway config is on the age-encrypted track, where integration said `create_`.
- The hermes config migration script arrived numbered 67, which on this tree is the rotate-logs
  LaunchAgent loader. `test/integration/hermes-route-status.sh` caught it: it refuses any mention of
  `run_after_67` in the route check next door, because a comment naming the wrong neighbour sends the
  next reader to the wrong file. The script took the free 59 slot instead, which satisfies the only
  ordering it needs (after chezmoi writes the config, before `run_after_68` reads it), and the route
  check now names it in full rather than by number.
- One thing to check rather than take on faith: the eight KeePassXC entry names in
  `dot_hermes/private_dot_env.tmpl`. They are ported verbatim from the tree that was applied to this
  machine, and no template on `main` references any of them, so nothing in this branch could cross-check
  them without opening the vault. They follow the repo's `Vendor :: Kind :: Detail` convention.


webdavis:
Reviewed all six items and the test that carries item 1.

Verified in the branch: the notifier now hands both tiers to relay with state, project and pane, no direct alerter call survives inside it, hue-pulse stays on the five minute tier, worktrunk is back to the herdr-aligned path, and effortLevel is a stable field. The shell producer is inside the progressive system rather than beside it.

The test is the deliverable and it earns that billing. Eleven mutations killed, including three that a weaker test would have missed: the dangling --pane that would silently drop --local-only and leak a desk-side notification to the phone, the threshold lowered under the documented 30s, and a direct alerter banner added beside relay. The equivalence control is the right call too, accepting either flag spelling while refusing the malformed one, and replacing the literal assertion that punished a legal form was the correct self-correction.

Two findings the suite produced rather than the author asserting them: the 67 numbering collision against main's log-rotation loader, caught by hermes-route-status and moved to the free 59 slot with the reason recorded, and the vault template proving itself wired through the secret-source-mode test that now pins .hermes/.env at 0600.

Flags accepted as flags: the eight KeePassXC entry names could not be cross-checked without opening the vault, so they are ported verbatim with the template's own caution comment intact, and the config-migrate drift against the committed age blob is integration's deployed design carried over rather than redesigned mid-port.

Approved.

PR #161, opened 2026-08-08
feat(pns): phone approvals and presence v2, with test purge and de-nix

webdavis:
## Context

The D1 cutover needs main to carry three things that were only local until now: pns v1 finished (phone
approvals through moshi-hook and a presence model that knows a phone in hand from a keyboard in use),
the test suite reduced to what the 2026-08-05 scope ruling keeps (behavior of tools this repo authors,
nothing else), and the last de-nix leftovers closed so a fresh checkout can run the gates without
guessing at tooling.

Every behavior here was verified live on dresden during two review sessions: approvals answered from
the phone, banners suppressed and fired per the presence rules, lights pulsed in both rooms, and the
Back Tap marker carried a notification that no other signal could have routed.

## Summary

- pns forwards blocking events to moshi-hook behind its presence gate: `hooks/moshi-gate.sh` is a pipe
  that execs moshi with stdin untouched, so the operator's approve or deny returns as the exit code
  (`dot_local/libexec/pns/hooks/`).
- Presence v2 in `dot_local/libexec/pns/helpers/presence.sh`: banner suppression requires pane focus,
  terminal frontmost and recent input all at once; a phone in hand is detected by mosh byte-rate or a
  Back Tap marker touched over forced-command SSH; three arbitration tiers make fresh keyboard input
  beat every phone signal.
- The banner click now works across workspaces: it resolves herdr's absolute path at notify time (the
  click runs in a bare launchd context where a bare name dies) and focuses workspace, then pane
  (`dot_local/libexec/pns/channels/executable_macos-banner.sh`).
- The unit suite is 188 bats tests after the scope purge (73 test files to 45, `test-system` suite
  deleted) and the osquery and pns conversions; every new assertion was mutation-verified, and one
  legacy test that posted a real macOS banner on every run now tests real absence.
- `Brewfile.dev` plus a `just setup` recipe pin the contributor toolchain, including mdformat at CI's
  exact versions; the nine test-tree references to the removed nix dev shell are corrected
  (`justfile`, `Brewfile.dev`, `CLAUDE.md`).

## How it was verified

- `just ship` exits 0: `just lint-check` (treefmt, ten formatters), `just test` (unit, integration,
  e2e, cargo), `just lint-actions-security` (actionlint plus zizmor, no findings).
- `just test-unit` alone: 188/188, seed-shuffled, no test over 150ms, and a recorder sweep proved no
  test reaches the real nettop, pgrep, lsappinfo or ioreg.
- Mutation sweeps on the new behavior: every suppression condition, both arbitration boundaries, the
  marker TTL, the caller-intent guards and the absolute-path click each go red under a named test when
  broken; two equivalent mutants are documented in place instead of hidden.
- Live drills on dresden, operator confirming each: a real Claude Code permission prompt answered from
  the phone; banner silence while watching the pane and banner delivery when another app was frontmost
  or the operator idle; the cross-workspace banner click landing on the source pane; phone cards
  delivered by the mosh byte-rate override and, with Moshi closed, by the Back Tap marker alone.

## Effect of merging

Merging updates main only. Nothing deploys by merging: chezmoi targets change at the next apply, and
the D1 cutover is that apply. dresden already runs these exact pns files through targeted applies made
during verification, so post-merge drift against the live machine is zero for the pns tree. The
`authorized_keys` line that lets the phone touch the presence marker was installed live and is not
managed by this repo yet; that gap is recorded in the plan notes.

## Review guide

Read `dot_local/libexec/pns/helpers/presence.sh` first: every decision the rest of the change relies
on lives there as pure functions. Then `hooks/moshi-gate.sh` (the exec-as-pipe contract and where its
one non-zero exit is correct), then `channels/executable_macos-banner.sh` (the three-condition
suppression and the click string), then the small override blocks in `executable_relay.sh` and
`hooks/executable_relay-agent.sh`. The bats files beside them state each behavior as a sentence and
are faster to read than the scripts.

The bulk of the diff is deletions from the test-scope purge and the .sh-to-bats conversions; they are
mechanical and the commit messages carry the reasoning. A second pair of eyes helps most on the
arbitration tiers in presence.sh (the fresh-input floor against the phone override) and on the
suppression ordering in the banner channel, since both encode operator rulings rather than code-forced
constraints.


.chezmoiscripts/run_after_46-bounce-moshi-hook-on-upgrade.sh.tmpl
webdavis: Delete these lines — doesn't need to be mentioned.

webdavis:
```suggestion
# STALENESS IS DECIDED BY INODE, NOT BY A VERSION STRING.
#
# WRONG WAY (version string):
#
#   moshi-hook --version   # reads the binary on disk
#   moshi-hook probe       # ALSO reads the binary on disk, not the daemon
#
# Both commands read the same file, so they always agree, even while the
# daemon is stale. Tested 2026-08-09: the daemon was serving 0.2.73, disk had
# 0.2.76, and probe said 0.2.76.
#
# RIGHT WAY (compare inodes):
#
#   lsof -p "$pid" | awk '$4 == "txt" { print $(NF-1) }'        # binary the daemon runs
#   stat -f '%i' "$(readlink -f /opt/homebrew/bin/moshi-hook)"  # binary on disk
#
# Brew doesn't edit the binary in place. An upgrade writes a new file, and a
# new file gets a new inode. The daemon keeps the file it opened at startup,
# no matter what lands on disk later. So if the two inodes match, the daemon
# is running today's binary. If they don't, brew swapped the file after the
# daemon started, and the bounce below fixes that.
```

The comment under the shebang needs to be updated as well.

webdavis: "Seam" sounds weird here and is ambiguous.

.chezmoiscripts/run_after_46-bounce-moshi-hook-on-upgrade.sh.tmpl
```diff
@@ -30,7 +30,82 @@ MOSHI_HOOK_BIN="${MOSHI_HOOK_BIN:-/opt/homebrew/bin/moshi-hook}"
...
+# --- the two ways the interposition breaks SILENTLY --------------------------
+#
+# pns registers itself as the harness hook and forwards to moshi-hook, so the
+# phone keeps its approve/deny round trip while pns keeps the presence gate.
+# Both halves of that arrangement can be undone without anything failing, which
+# is why they are re-asserted here rather than trusted: this script already runs
+# after every upgrade, which is when both happen.
+#
+# Seams, so a test drives this without touching the operator's live files or
+# the real binary. Same names the hooks use.
```
webdavis: Make this more readable. Currently, it's gibberish.

dot_local/libexec/pns/channels/executable_hermes.sh
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env bash
```
webdavis, line 1: Comments in this file need to be clearer and simpler.

dot_local/libexec/pns/channels/executable_hue-pulse.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
```
webdavis, line 1: Comments in this file need to be clearer and simpler.

dot_local/libexec/pns/channels/executable_macos-banner.sh
```diff
@@ -0,0 +1,109 @@
+#!/usr/bin/env bash
```
webdavis, line 1: Comments in this file need to be clearer and simpler.

dot_local/libexec/pns/channels/executable_moshi.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
```
webdavis, line 1: Comments in this file need to be clearer and simpler.

dot_local/libexec/pns/executable_relay.sh
webdavis, line 1: Comments in this file need to be clearer and simpler.

dot_bashrc.tmpl
```diff
@@ -439,11 +439,25 @@ if [[ $- == *i* ]]; then
   # ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
   # Uses bash-preexec (sourced above) instead of a naked DEBUG trap, which would
   # clobber atuin's recording.
+  #
+  # THE TIMER IS PS0, NOT PREEXEC. bash-preexec never fires preexec for the
```
webdavis: This comment needs to be clearer and simpler.

dot_bashrc.tmpl
```diff
@@ -452,6 +466,10 @@ if [[ $- == *i* ]]; then
     [[ -z $__cmd_notify_start ]] && return
     local elapsed=$((SECONDS - __cmd_notify_start))
     __cmd_notify_start=""
+    # First command of a session: preexec never ran, so the name is empty and
```
webdavis: This comment needs to be clearer and simpler.

dot_bashrc.tmpl
```diff
@@ -464,14 +482,21 @@ if [[ $- == *i* ]]; then
     # relay producer like the Claude and Codex hooks and the weekly jobs, rather
     # than a second notifier beside it: relay decides presence (at the desk,
     # banner only; away, add the phone push), writes the Discord paper trail, and
-    # makes the banner clickable, which is what carries the pane. --local-only
-    # keeps the first tier to the banner; the second tier fans out and pulses the
-    # lights. test/integration/bashrc-long-command-relay.sh pins that contract.
+    # makes the banner clickable, which is what carries the pane. Both tiers go
```
webdavis: This comment needs to be clearer and simpler.

dot_bashrc.tmpl
```diff
@@ -464,14 +482,21 @@ if [[ $- == *i* ]]; then
...
+    # template to assert its text, which is declaration checking rather than tool
+    # behavior, so it went with the 2026-08-05 scope ruling. Change a threshold
+    # here and only a real long command will tell you.
     if ((elapsed >= 300)); then
-      ("$HOME/.local/libexec/pns/hue-pulse.sh" "$exit_code" >/dev/null 2>&1 &)
+      ("$HOME/.local/libexec/pns/channels/hue-pulse.sh" "$exit_code" >/dev/null 2>&1 &)
       ("$HOME/.local/libexec/pns/relay.sh" --agent shell --state "$state" --project "${PWD##*/}" --detail "$cmd ($dur)" --pane "${HERDR_PANE_ID:-}" >/dev/null 2>&1 &)
     elif ((elapsed >= 30)); then
-      ("$HOME/.local/libexec/pns/relay.sh" --local-only --agent shell --state "$state" --project "${PWD##*/}" --detail "$cmd ($dur)" --pane "${HERDR_PANE_ID:-}" >/dev/null 2>&1 &)
+      # NOT --local-only (operator ruling 2026-08-06): away means mobile, and
```
webdavis: This comment needs to be clearer and simpler.

justfile
```diff
@@ -75,9 +75,12 @@ apply-no-auth:
 # Tests live in suites by DESIGN: test/unit (single component, stub-driven, no
 # flows, no sleeps; FAST is the admission rule), test/integration and test/e2e
 # (bats coverage for the long-lived bash tools, speed-gated like everything
-# else since the 2026-08-05 purge), test/test-system (the runner's own tests).
-# The pre-commit hook runs `just test-unit` only; the pre-push hook runs no
-# suite at all (lint drift only); CI and `just ship` run `just test`.
+# else since the 2026-08-05 purge). A fourth camp does NOT live under test/,
```
webdavis: This comment needs to be clearer and simpler.

